Issue #311, opened 2023-03-30
Ligretto loading screen remove scrolls


Themezv:
Current:
<img width="1728" alt="image" src="https://user-images.githubusercontent.com/12786682/228888600-983bcd7e-112e-4502-9b00-96dc1957c858.png">

NewtonFernandis:
Hi @Themezv , Could you please assign this to me?

PR #312, opened 2023-03-31
Removed scrollbar from loading screen

NewtonFernandis:
Fixes issue Ligretto loading screen remove scrolls #311

### DIFF
--- a/apps/ligretto-frontend/src/components/screens/LoaderScreen/LoaderScreen.tsx
+++ b/apps/ligretto-frontend/src/components/screens/LoaderScreen/LoaderScreen.tsx
@@ -12,7 +12,16 @@ export const LoaderScreen: React.FC = () => {
 
   return (
     <MainLayout>
-      <Box display="flex" flexDirection="column" justify-items="center" align-items="center" height="100%">
+      <Box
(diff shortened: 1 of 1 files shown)


apps/ligretto-frontend/src/components/screens/LoaderScreen/LoaderScreen.tsx
```diff
@@ -12,7 +12,16 @@ export const LoaderScreen: React.FC = () => {
 
   return (
     <MainLayout>
-      <Box display="flex" flexDirection="column" justify-items="center" align-items="center" height="100%">
+      <Box
+        style={{
```
Themezv, line 16: Unfortunately we don't use style property (and sx) for styling.
As I know Box accept overflow property as usual prop, so we can try that: 
```suggestion
        overflow="hidden"
```

Themezv: @NewtonFernandis Could you fix it?

Themezv:
Fixed in https://github.com/MemeBattle/monorepo/pull/351

PR #638, opened 2026-08-11
refactor(ligretto): centralize card focus management

### LINKED ISSUE #636: Centralize card focus management and card interactions
## Goal

Introduce a dedicated frontend module that owns focused-card state (currently `selectedCardIndex`) and centralizes every card interaction so clicks and hotkeys follow the same rules.

This supersedes the isolated stack-rotation bugfix: the stale selection is one symptom of focus ownership being split across Redux game listeners, UI containers, and the hotkey hook.

## Problems to solve

- Focus state is stored as a low-level row index/sentinel inside the game slice.
- Focus transitions are split between game listeners and React components.
- Escape directly mutates focus while other hotkeys dispatch game actions.
- Row cards and the open stack card duplicate selected/dimmed rendering logic.
- Outside-click behavior is registered per card instead of at one interaction boundary.
- Pointer and keyboard paths can diverge as new interactions are added.

## Requirements

Create a card-focus module that:

- owns semantic focus state for row cards and the open stack card;
- owns focus, toggle, dismiss, validation, and post-interaction clearing rules;
- exposes hooks/components for card UI and hotkey integration;
- accepts semantic interactions for row cards, open/closed stack cards, the Ligretto deck, playground cards, and dismissal;
- translates those interactions into the existing shared game actions without backend or protocol changes;
- gives pointer and keyboard input the same behavior;
- clears focus when stack rotation or another transition invalidates the focused card;
- preserves immediate-play behavior in automatic placement mode and for cards that must be played immediately.

## Migration

Migrate these consumers away from direct `selectedCardIndex` and `tap*Action` ownership:

- `PlayerRowCardsContainer`
- `PlayerCardsStack`
- `CardsPanelContainer/usePanelHotkeys`
- `LigrettoDeckContainer`
- `PlaygroundContainer`
- the focus-related portions of `ducks/game/slice.ts`, `selectors.ts`, and `listeners.ts`

Remove `selectedCardIndex`, `setSelectedCardIndexAction`, and compatibility code after all consumers use the new module.

## Acceptance criteria

- Clicking a card and using its hotkey produce the same semantic interaction.
- Row-card and open-stack focus can be selected, toggled, rendered, and dismissed through module APIs only.
- Space/click stack rotation clears focus and rotates the stack exactly once.
- Escape and outside-pointer dismissal clear focus without emitting a game command.
- Playground placement consumes the correct focused card and clears focus.
- Invalid focus is reconciled after game updates.
- Automatic placement mode and Ligretto deck behavior do not regress.
- Production code has no remaining references to the old selected-card-index API.
- Frontend tests, type checks, lint, and formatting checks pass.

## Implementation plan

The architecture, phased migration, test matrix, risks, and commit sequence are documented in [`docs/card-focus-management-plan.md`](https://github.com/MemeBattle/monorepo/blob/fix/issue-636-clear-selection/docs/card-focus-management-plan.md) and submitted for review in #638.

memebattle-dev:
## Summary

- add a React Context card-focus feature with provider-owned state and outside-click handling;
- keep hotkeys in `usePanelHotkeys` and route focus operations through the shared hook;
- migrate row/open-stack focus, stack rotation, and playground placement to the shared focus API;
- remove `selectedCardIndex` and related Redux state/listeners;
- retain existing Redux game actions and backend payloads;
- add card-focus behavior tests for toggle/transfer, identity changes, outside clicks, and disabled state, plus playground dismissal coverage.

## Validation

- `tsc --noEmit`
- `vitest run` — 34 tests passed
- `oxlint apps/ligretto-frontend/src`
- `oxfmt --check` on all changed files
- `vite build`
- `git diff --check`

Closes #636


apps/ligretto-frontend/src/ducks/game/listeners.spec.ts
```diff
@@ -0,0 +1,48 @@
+// @vitest-environment jsdom
```
memebattle-dev: Is this necessary?

memebattle-dev: Yes. I verified this by removing the directive: the suite then fails during module import with `ReferenceError: window is not defined` from `@memebattle/auth-front` before any tests are collected. The listener module's import graph reaches browser-dependent application code, so jsdom is required for this focused listener test. Restoring the directive makes both regression cases pass.

apps/ligretto-frontend/src/ducks/game/listeners.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+  const store = configureStore({
+    reducer: { game: gameReducer },
+    middleware: getDefaultMiddleware =>
+      getDefaultMiddleware({ serializableCheck: false, thunk: false }).prepend(listenerMiddleware.middleware).concat(recordActions),
+  })
+
+  return { store, dispatchedActions }
+}
+
+describe('game listeners', () => {
```
Themezv: I think this type of test isn’t very useful.

We prefer testing the application’s behavior from the user’s perspective

In this case it would be better to use RTL to activate the card and then press Space or click the face-down card in the stack.

Themezv (changes requested):

apps/ligretto-frontend/src/ducks/game/listeners.spec.ts
memebattle-dev: Addressed in 7554148. I replaced the listener-level test with React Testing Library coverage that selects a row card through the rendered UI, then rotates the stack through both Space and a click on the face-down card. Both cases assert that the selection clears and that the stack rotation action is dispatched exactly once with the expected game ID. The frontend test suite, typecheck, lint, and formatting checks pass.

docs/card-focus-management-plan.md
```diff
@@ -0,0 +1,251 @@
...
+
+```ts
+export type CardFocusTarget = { type: 'row'; index: number } | { type: 'stack-open' }
+
+export type CardInteraction =
+  | { type: 'row-card'; index: number }
+  | { type: 'stack-open-card' }
+  | { type: 'stack-deck' }
+  | { type: 'ligretto-deck' }
+  | { type: 'playground-card'; deckPosition: number }
```
StepanDyubin: useless, the same behaviour as dismiss, remove them

docs/card-focus-management-plan.md
```diff
@@ -0,0 +1,251 @@
...
+Only cards that can participate in manual placement are focus targets. The closed stack, Ligretto deck, and playground remain interaction targets but are not focusable cards.
+
+Suggested module boundary:
+
+```text
+features/cardFocus/
+├── index.ts
+├── model/
+│   ├── types.ts
+│   ├── slice.ts
```
StepanDyubin: Do not use redux that feature. Use react context

docs/card-focus-management-plan.md
```diff
@@ -0,0 +1,251 @@
...
+| Ligretto deck                              | Draw/play according to existing behavior | Preserve existing behavior; clear focus if the resulting game transition invalidates it |
+| Playground card                            | Existing no-op/play behavior             | If a card is focused, place it and clear focus after a successful command is emitted    |
+| Escape                                     | No-op                                    | Clear focus                                                                             |
+| Outside pointer interaction                | No focus change                          | Clear focus once through the provider boundary, not once per card                       |
+| Focused card disappears or becomes invalid | N/A                                      | Clear focus during state reconciliation                                                 |
+
+Click and hotkey paths must produce the same semantic `CardInteraction`. UI components must not duplicate transition rules.
```
StepanDyubin: We also should clear the focus if card component changes somehow (re-render with new value).
and if user click outside any card. Add the parts in the plan

StepanDyubin (commented):
@memebattle-dev Do corrections

docs/card-focus-management-plan.md
```diff
@@ -0,0 +1,275 @@
...
+
+**Files:**
+
+- Modify `apps/ligretto-frontend/src/features/player/ui/CardsPanelContainer/usePanelHotkeys.ts`.
+- Modify `apps/ligretto-frontend/src/features/player/ui/PlayerCardsStack/PlayerCardsStack.tsx`.
+- Modify `apps/ligretto-frontend/src/features/playground/ui/PlaygroundContainer.tsx` or the placement integration point.
+
+**Steps:**
+
+1. Replace the Escape Redux focus action with `clearCardFocus()`.
```
StepanDyubin: Where could that and useClearCardFocus hook be usefull?

memebattle-dev: `clearCardFocus` is the context operation; `useClearCardFocus` would only expose that operation to non-card consumers. The intended consumers are Escape in `usePanelHotkeys`, closed-stack rotation (click and Space), successful playground placement, and switching out of manual placement mode. However, a dedicated hook for one context method is unnecessary—we can expose it through a general context hook (for example, `useCardFocusContext`) and remove `useClearCardFocus` from the plan.

memebattle-dev: Addressed in 19083d6. I removed the dedicated `useClearCardFocus` hook from the plan and replaced it with a general `useCardFocusContext` API. The plan now states that `clearCardFocus` is used by non-card integrations such as Escape handling, closed-stack rotation, and successful playground placement, while `focusedCard` supports placement command construction.

docs/card-focus-management-plan.md
```diff
@@ -0,0 +1,277 @@
...
+Suggested module boundary:
+
+```text
+features/cardFocus/
+├── index.ts
+├── model/
+│   └── types.ts
+└── ui/
+    ├── CardFocusContext.ts
+    ├── CardFocusProvider.tsx
```
StepanDyubin, line 41: Lets handle both: hotkeys and dismiss focus by clicking outside in the provider. And do not expose nothing but provider and hook (or maybe some wrapper) from the module.

Add more details about outside click handling

apps/ligretto-frontend/src/features/player/ui/CardsPanelContainer/usePanelHotkeys.ts
```diff
@@ -1,52 +0,0 @@
...
-import {
-  Hotkey,
-  setSelectedCardIndexAction,
-  tapCardAction,
-  tapLigrettoDeckCardAction,
-  tapStackDeckCardAction,
-  tapStackOpenDeckCardAction,
-} from '#ducks/game'
-
-export const usePanelHotkeys = ({ enabled }: { enabled?: boolean } = { enabled: true }) => {
```
StepanDyubin, line 12: Keep the hook.

it should be used inside CardsPanelContainer, remove that logic from provider.
usePanelHotkeys should call focus actions here if dnd is enabled

apps/ligretto-frontend/src/features/player/ui/LigrettoPack/LigrettoPack.tsx
```diff
@@ -41,6 +43,7 @@ export const LigrettoPack = ({
             onClick={onLigrettoDeckCardClick}
             isHighlighted={isHighlighted}
             isDisabled={isDisabled}
+            dataCardFocusElement={markCardForFocus}
```
StepanDyubin: On click to ligretto deck, card should just appear in row. So it is not focusable, do not pass it through

apps/ligretto-frontend/src/features/cardFocus/ui/useCardFocus.ts
```diff
@@ -0,0 +1,77 @@
+import { useCallback, useContext, useEffect, useMemo, useRef } from 'react'
+
+import { CardFocusContext, isSameCardFocusTarget, type CardFocusTarget, type CardIdentity } from './CardFocusContext'
+
+interface UseCardFocusOptions {
+  target: CardFocusTarget
+  identity: CardIdentity
```
StepanDyubin: Do not pass identity.
Pass just deps (as it is in useEffect). and clear focus on change

docs/card-focus-management-plan.md
memebattle-dev: Addressed in d8440cd (with the empty-playground dismissal regression corrected in 6d08645). `CardFocusProvider` now owns all card hotkeys and the single document-level outside-card click listener; the module exports only `CardFocusProvider` and `useCardFocus`. The detailed plan and the existing issue plan comment were updated to describe event ordering, card-root markers, cleanup, and dismissal behavior.

apps/ligretto-frontend/src/entities/card/ui/Card/Card.tsx
StepanDyubin: remove onClickOutside, provider manages that

apps/ligretto-frontend/src/features/cardFocus/ui/useCardFocus.ts
```diff
@@ -0,0 +1,77 @@
...
+  const clearFocus = useCallback(() => setFocusedCard(undefined), [setFocusedCard])
+
+  return options
+    ? {
+        focusedCard,
+        isFocused,
+        isDimmed: !!focusedCard && !isFocused,
+        toggleFocus,
+        clearFocus,
+        cardFocusProps: { dataCardFocusElement: true as const },
```
StepanDyubin: Remove that prop from all components.

The only cards that uses useCardFocus should be focusable - row card and open stack. they should define that itself.

Let's create "FocusableCard" wrapper that uses `useCardFocus` and add html block (with data attribute).

So as a result we have to just wrap each card component (that should be focusable) + provider

apps/ligretto-frontend/src/features/player/ui/CardsPanelContainer/usePanelHotkeys.ts
memebattle-dev: Addressed in fc1eb9d. `usePanelHotkeys` is restored in `CardsPanelContainer`, gated by manual placement and active-game state, and now calls the card-focus hook for focus actions while keeping Redux game commands unchanged.

apps/ligretto-frontend/src/features/player/ui/LigrettoPack/LigrettoPack.tsx
memebattle-dev: Addressed in fc1eb9d. The Ligretto deck is no longer marked or wrapped as focusable; clicking it only dispatches the existing deck action.

apps/ligretto-frontend/src/features/cardFocus/ui/useCardFocus.ts
memebattle-dev: Addressed in fc1eb9d. `useCardFocus` now accepts a dependency list and clears the focused target through the effect cleanup whenever those dependencies change, without a separate identity object.

apps/ligretto-frontend/src/entities/card/ui/Card/Card.tsx
memebattle-dev: Addressed in fc1eb9d. `onClickOutside` and the per-card outside-click hook were removed from `Card`; outside dismissal remains provider-owned.

apps/ligretto-frontend/src/features/cardFocus/ui/useCardFocus.ts
memebattle-dev: Addressed in fc1eb9d. Focus props were removed from shared card components. The new exported `FocusableCard` wrapper owns `useCardFocus` and the private data marker, and only row cards and the open-stack card are wrapped.

apps/ligretto-frontend/src/entities/card/ui/CardPlace/CardPlace.tsx
```diff
@@ -51,8 +51,8 @@ const StyledCard = styled('div')<{ size: CardPlaceSize }>(({ size, theme }) => (
   },
 }))
 
-export const CardPlace = ({ children, size = 'medium', ref, dataTestId }: PropsWithChildren<CardPlaceProps>) => (
-  <StyledCardPlace ref={ref} size={size} data-test-id={dataTestId}>
+export const CardPlace = ({ children, size = 'medium', ref, dataTestId, onClick }: PropsWithChildren<CardPlaceProps & { onClick?: () => void }>) => (
+  <StyledCardPlace ref={ref} size={size} data-test-id={dataTestId} onClick={onClick}>
```
StepanDyubin: Remove that change

apps/ligretto-frontend/src/features/player/ui/PlayerCardsStack/PlayerCardsStack.tsx
StepanDyubin: Remove CardsStack component, inline inner jsx.
Create "PlayerStackOpenCard" and "PlayerStackDeck" components. They are wrappers only for Card (Not CardPlace, badge ...) and incapsulates related logic. Use them here. 

make Card to pass through data-atributes (for detecting outside click) - just spread the rest props

apps/ligretto-frontend/src/features/cardFocus/ui/FocusableCard.tsx
```diff
@@ -0,0 +1,22 @@
...
+  children: (focus: {
+    focusedCard: CardFocusTarget | undefined
+    isFocused: boolean
+    isDimmed: boolean
+    toggleFocus: () => void
+    clearFocus: () => void
+  }) => ReactNode
+}
+
+export const FocusableCard = ({ target, deps, children }: FocusableCardProps) => {
```
StepanDyubin: Remove FocusableCard.

We should have now PlayerRowCard and PlayerStackOpenCard that can use the hook inside

apps/ligretto-frontend/src/features/cardFocus/ui/CardFocusContext.ts
```diff
@@ -0,0 +1,18 @@
+import { createContext, type Dispatch, type SetStateAction } from 'react'
+
+export type CardFocusTarget = { type: 'row'; index: number } | { type: 'stack-open' }
```
StepanDyubin: Let remove CardFocusTarget as it is now.

Make it be just a string key.
PlayerRowCard's key = row.[index]
PlayerOpenStackCard's key = 'stack-open'

the components pass a key themselfs.

Key is easy to compare, and it also can be used as `deps` to clear focus

apps/ligretto-frontend/src/features/cardFocus/ui/useCardFocus.ts
```diff
@@ -0,0 +1,71 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [setFocusedCard, target, ...deps],
+  )
+
+  const toggleFocus = useCallback(() => {
+    if (target) {
+      setFocusedCard(current => (isSameCardFocusTarget(current, target) ? undefined : target))
+    }
+  }, [setFocusedCard, target])
+  const toggleTargetFocus = useCallback(
```
StepanDyubin: keep the only methods in use

apps/ligretto-frontend/src/entities/card/ui/CardPlace/CardPlace.tsx
memebattle-dev: Addressed in 7b35b5c. The `CardPlace` click prop and forwarding were removed.

apps/ligretto-frontend/src/features/player/ui/PlayerCardsStack/PlayerCardsStack.tsx
memebattle-dev: Addressed in 7b35b5c. The generic `CardsStack` component was removed and its layout was inlined. `PlayerStackOpenCard` and `PlayerStackDeck` now wrap `Card` only and encapsulate their interaction logic; `Card` forwards data attributes via the remaining props.

apps/ligretto-frontend/src/features/cardFocus/ui/FocusableCard.tsx
memebattle-dev: Addressed in 7b35b5c. `FocusableCard` was removed. `PlayerRowCard` and `PlayerStackOpenCard` now call `useCardFocus` directly and mark the actual `Card` element.

apps/ligretto-frontend/src/features/cardFocus/ui/CardFocusContext.ts
memebattle-dev: Addressed in 7b35b5c. Focus targets are now string keys (`row.<index>` and `stack-open`), with the key included in the hook cleanup dependencies.

apps/ligretto-frontend/src/features/cardFocus/ui/useCardFocus.ts
memebattle-dev: Addressed in 7b35b5c. The keyed hook overload now returns only `isFocused`, `isDimmed`, `toggleFocus`, and `clearFocus`; the unkeyed overload retains the operations used by hotkeys and playground placement.

apps/ligretto-frontend/src/features/playground/ui/Playground.spec.tsx
```diff
@@ -0,0 +1,19 @@
+// @vitest-environment jsdom
+
+import { fireEvent, render } from '@testing-library/react'
+import { describe, expect, it, vi } from 'vitest'
+
+import { Playground } from './Playground'
+
+describe('Playground', () => {
```
StepanDyubin: Remove that test

apps/ligretto-frontend/src/features/cardFocus/ui/useCardFocus.ts
```diff
@@ -0,0 +1,64 @@
+import { useCallback, useContext, useEffect, type DependencyList } from 'react'
+
+import { CardFocusContext, type CardFocusKey } from './CardFocusContext'
+
+interface UseCardFocusOptions {
+  focusKey: CardFocusKey
+  deps: DependencyList
+}
```
StepanDyubin: Lets make options be
{
   type: 'open-stack',
   card: { color, value }
} | {
   type: 'row',
   index,
   card: { color, value }
}

It should be the only option for the hook
`useCardFocus({ type: 'open-stack', card })`

And form the key to clear focus as type.[mayby index].cardColor.CardValue

apps/ligretto-frontend/src/features/playground/ui/Playground.spec.tsx
memebattle-dev: Removed `Playground.spec.tsx` as requested in a9118bb.

apps/ligretto-frontend/src/features/cardFocus/ui/useCardFocus.ts
memebattle-dev: Updated `useCardFocus` in a9118bb. It now accepts only `{ type: "open-stack", card } | { type: "row", index, card }`; the internal key is derived from type, optional index, color, and value. All callers and the plan were migrated.

apps/ligretto-frontend/src/features/cardFocus/ui/useCardFocus.ts
```diff
@@ -0,0 +1,76 @@
...
+  clearFocus: () => void
+}
+export function useCardFocus(options?: CardFocusOptions) {
+  const context = useContext(CardFocusContext)
+  if (!context) {
+    throw new Error('useCardFocus must be used within CardFocusProvider')
+  }
+
+  const { focusedCard, setFocusedCard } = context
+  const focusKey = options && getCardFocusKey(options)
```
StepanDyubin: Use focusKey only internally for useEffect as comparing targets,
Tho hook should return the same type as option CardFocusOptions (rename it)

memebattle-dev: Updated in 2c8e7f0. The hook/context now store and return the structured `CardFocusOptions` value. The derived `focusKey` remains private to `useCardFocus` and is used only for the effect dependency and cleanup comparison. Playground placement now consumes the structured target directly.

apps/ligretto-frontend/src/features/cardFocus/ui/useCardFocus.ts
```diff
@@ -0,0 +1,70 @@
...
+    : `${target.type}.${target.card.color}.${target.card.value}`
+
+const isSameCardFocusTarget = (left: CardFocusOptions | undefined, right: CardFocusOptions | undefined) =>
+  left === right ||
+  (!!left &&
+    !!right &&
+    left.type === right.type &&
+    left.card.color === right.card.color &&
+    left.card.value === right.card.value &&
+    (left.type === 'open-stack' || (right.type === 'row' && left.index === right.index)))
```
StepanDyubin: just compare keys here

apps/ligretto-frontend/src/features/player/ui/CardsPanelContainer/usePanelHotkeys.ts
```diff
@@ -1,49 +1,71 @@
...
+            dispatch(tapStackOpenDeckCardAction())
+          } else {
+            toggleFocus({ type: 'open-stack', card: stackOpenDeckCard })
+          }
+          break
         case Hotkey.space:
+          clearFocus()
           dispatch(tapStackDeckCardAction())
           break
-        case Hotkey.q:
```
StepanDyubin, line 24: Keep that part hardcoded as it was

case q:
   if (value ...) {
      toggleFocus(...)
   }

case w: 
    if (value ...) {
    toggleFocus(...)

apps/ligretto-frontend/src/features/cardFocus/ui/useCardFocus.ts
memebattle-dev: Simplified in a1fcc53: `isSameCardFocusTarget` now compares the derived keys.

apps/ligretto-frontend/src/features/player/ui/CardsPanelContainer/usePanelHotkeys.ts
memebattle-dev: Restored explicit hardcoded Q/W/E/R/T switch cases in a1fcc53, with each case applying the value/immediate-play check and toggling the corresponding structured focus target.

apps/ligretto-frontend/src/features/cardFocus/ui/useCardFocus.ts
```diff
@@ -0,0 +1,64 @@
...
+
+const isSameCardFocusTarget = (left: CardFocusOptions | undefined, right: CardFocusOptions | undefined) =>
+  left === right || (!!left && !!right && getCardFocusKey(left) === getCardFocusKey(right))
+
+export function useCardFocus(): {
+  focusedCard: CardFocusOptions | undefined
+  clearFocus: () => void
+  toggleFocus: (target: CardFocusOptions) => void
+}
+export function useCardFocus(target: CardFocusOptions): {
```
StepanDyubin: 1. Let's remove `card` field from CardFocusOptions.
2. add second arg to the hook - deps - array of deps where focus is reseting.

memebattle-dev: Updated in d762ddc: removed `card` from `CardFocusOptions` and changed the card overload to `useCardFocus(target, deps)`. Row/open-stack components pass `[card.color, card.value]`, so focus cleanup follows the explicit identity dependencies while hotkeys use location-only targets.

apps/ligretto-frontend/src/features/playground/ui/PlaygroundContainer.tsx
```diff
@@ -1,30 +1,50 @@
...
+      }
+
+      if (focusedCard.type === 'open-stack') {
+        dispatch(putCardFromStackOpenDeck({ gameId, playgroundDeckIndex }))
+      } else if (focusedCard.type === 'row') {
+        dispatch(putCardAction({ cardIndex: focusedCard.index, gameId, playgroundDeckIndex }))
+      } else {
+        return
+      }
+      clearFocus()
```
StepanDyubin: Do not clear focus, it will happen automatically when focused card is unmounted

apps/ligretto-frontend/src/features/playground/ui/PlaygroundContainer.tsx
```diff
@@ -1,30 +1,50 @@
...
+    },
+    [clearFocus, dispatch, focusedCard, gameId, isDndEnabled],
+  )
+
+  const handlePlaygroundClick = useCallback(
```
StepanDyubin: useless, remove

apps/ligretto-frontend/src/features/player/ui/PlayerRowCardsContainer/PlayerRowCardsContainer.tsx
```diff
@@ -1,58 +1,52 @@
...
+const PlayerRowCard = ({ card, index, isDndEnabled, hotkey }: PlayerRowCardProps) => {
+  const dispatch = useDispatch()
+  const { isFocused, isDimmed, toggleFocus, clearFocus } = useCardFocus({ type: 'row', index }, [card.color, card.value])
+  const onCardClick = () => {
+    if (isDndEnabled && card.value !== 1) {
+      toggleFocus()
+    } else {
+      clearFocus()
       dispatch(tapCardAction({ cardIndex: index }))
-    },
```
StepanDyubin, line 27: always toggle focus, clear is called when card unmounts

apps/ligretto-frontend/src/features/player/ui/PlayerCardsStack/PlayerStackOpenCard.tsx
```diff
@@ -0,0 +1,32 @@
...
+      type: 'open-stack',
+    },
+    [card.color, card.value],
+  )
+
+  const handleClick = () => {
+    if (isDndEnabled && card.value !== 1) {
+      toggleFocus()
+    } else {
+      clearFocus()
```
StepanDyubin: always toggle focus, clear is called when card unmounts

apps/ligretto-frontend/src/features/playground/ui/PlaygroundContainer.tsx
memebattle-dev: Addressed in 6e855bc: playground placement no longer clears focus explicitly; the focused card lifecycle now owns cleanup.

apps/ligretto-frontend/src/features/playground/ui/PlaygroundContainer.tsx
memebattle-dev: Removed the redundant wrapper callback in 6e855bc and pass the placement handler directly to `Playground`.

apps/ligretto-frontend/src/features/player/ui/PlayerRowCardsContainer/PlayerRowCardsContainer.tsx
memebattle-dev: Addressed in 6e855bc: row-card clicks now always toggle focus before the immediate-play branch; lifecycle cleanup handles card replacement/unmount.

apps/ligretto-frontend/src/features/player/ui/PlayerCardsStack/PlayerStackOpenCard.tsx
memebattle-dev: Addressed in 6e855bc: open-stack clicks now always toggle focus before dispatching the immediate game action when applicable.

apps/ligretto-frontend/src/features/player/ui/CardsPanelContainer/usePanelHotkeys.ts
```diff
@@ -1,49 +1,109 @@
...
       switch (handler.hotkey) {
-        case Hotkey.space:
-          dispatch(tapStackDeckCardAction())
-          break
-        case Hotkey.q:
-          dispatch(tapCardAction({ cardIndex: 0 }))
+        case Hotkey.q: {
+          const card = playerCards?.[0]
+          if (!card) {
+            clearFocus()
```
StepanDyubin: We should not rely on game state while managing focus.

Let's improve the focus system, so each card when using useCardFocus should register itself in provider's state.
And if we call for exmpl `toggleFocus({ type: 'row', index: 0 })` if there is no registred card in the moment, we just clearFocus (internally)

@memebattle-dev write some short plan how to do it in the comment's reply first

memebattle-dev: Short plan:

1. Add an internal provider-owned registry keyed by `CardFocusOptions` (`row.<index>` / `open-stack`). The context will expose private register/unregister operations; they will not be part of the feature public exports.
2. Make `useCardFocus(target, deps)` register its card on mount, refresh the registration when `target`/`deps` change, and unregister on cleanup. Unregistering the focused target will clear focus.
3. Store the card-owned activation metadata/callback in the registration, so keyboard activation uses the rendered card as the source of truth (including immediate-play/value-1 behavior) instead of reading cards from Redux selectors.
4. Change provider focus operations to resolve the target through the registry first. If `toggleFocus(target)` has no current registration, it will clear focus; otherwise it will execute the registered card interaction/toggle rules.
5. Remove `playerCardsSelector` and `playerStackOpenDeckCardsSelector` from `usePanelHotkeys`; keep the explicit Q/W/E/R/T/X cases, but route each target through the provider registry. Space/L/Escape keep their existing command/clear behavior.
6. Add tests for registration/unregistration, missing-target clearing, identity replacement, immediate-play routing, and hotkeys without game-state card selectors; then run lint, typecheck, tests, and build.

memebattle-dev: Implemented in cd209bc. Cards now register their target and activation behavior in `CardFocusProvider`; missing targets clear focus, unregistration clears matching focus, and non-focusable/immediate-play cards route through their registered callback. `usePanelHotkeys` no longer reads card state from Redux and only sends explicit Q/W/E/R/T/X targets to the focus feature. Added registry, missing-target, activation, and unmount tests.

apps/ligretto-frontend/src/features/cardFocus/ui/useCardFocus.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export function useCardFocus(): {
+  focusedCard: CardFocusOptions | undefined
+  clearFocus: () => void
+  toggleFocus: (target: CardFocusOptions) => void
+}
+export function useCardFocus(
+  target: CardFocusOptions,
+  deps: DependencyList,
+  registration?: Partial<CardFocusRegistration>,
```
StepanDyubin: Remove registration option

all components that uses useCardFocus with target is focusable.
onActivate is also useless, keep logic in handlers

memebattle-dev: Updated in 3979d87. Removed the registration option, `canFocus`, and `onActivate`. Every targeted `useCardFocus` call now registers only its focusable target. Immediate-play/game-action logic remains in the row/open-stack click handlers, while generic target toggles still clear focus when no card is registered.

StepanDyubin (approved):

PR #686, opened 2026-08-31
feat(ligretto): unify card interactions and drag and drop

### LINKED ISSUE #685: Add drag-and-drop card placement
## Feature description

Add direct drag-and-drop placement for playable Ligretto cards in the active game UI.

Players should be able to drag a card from their row or open stack onto a valid playground deck instead of using only the existing select-then-click interaction.

## Expected behavior

- Row cards and the visible open-stack card can be dragged.
- Playground decks act as drop targets.
- Valid targets are visually highlighted while dragging.
- Dropping on a valid deck dispatches the existing placement command with the explicit `playgroundDeckIndex`.
- Dropping on an invalid target, outside the playground, or cancelling the gesture leaves the game state unchanged and returns the card to its origin.
- Dragging a value-1 card onto an empty deck places it on the chosen deck; its existing click/hotkey automatic-placement behavior remains unchanged.
- Card focus is cleared after successful placement and remains consistent after cancelled/invalid drops.
- Mouse, touch, and pointer input are supported.

## Compatibility requirements

Drag-and-drop must be an additional interaction, not a replacement:

- preserve click-to-focus followed by playground click;
- preserve card hotkeys and keyboard accessibility;
- preserve automatic click/hotkey placement for value-1 cards;
- spectators and inactive players cannot drag cards;
- onboarding behavior should not change unless it explicitly adopts the shared drag-and-drop components in a separate scope.

## Implementation considerations

- Reuse the current structured card targets (`row/index` and `open-stack`) and existing placement actions.
- Keep authoritative move validation on the gameplay backend.
- Avoid duplicating placement-command construction between click and drop paths.
- Provide clear drag preview, valid/invalid target states, and accessible non-drag alternatives.
- Prefer pointer-based behavior that works consistently on desktop and mobile; evaluate a maintained React drag-and-drop library before introducing custom global listeners.

## Acceptance criteria

- [ ] A row card can be dragged to a valid playground deck.
- [ ] The open-stack card can be dragged to a valid playground deck.
- [ ] A value-1 card can be dropped onto a chosen empty deck while click/hotkey still auto-places it.
- [ ] Invalid and cancelled drops dispatch no gameplay command.
- [ ] Valid drop targets receive visual feedback.
- [ ] Drag state cleans up when the card changes, unmounts, the game ends, or the player becomes inactive.
- [ ] Click, hotkey, and keyboard-accessible placement continue to work.
- [ ] Behavior is covered by component/integration tests and Playwright tests for mouse and touch/pointer flows.

## Context

Follow-up to #635 and the centralized card-focus work from #636.

memebattle-dev:
## Summary

- Unify card selection, mouse/touch drag-and-drop and hotkeys in `cardInteraction`, with an `idle` / `focused` / `dragging` reducer.
- Use standard dnd-kit sensors: mouse movement beyond 6 px and touch movement beyond 8 px. Stationary touch holds preserve normal click selection. Hotkeys remain available during a drag without custom cancellation.
- Keep provider-owned input gating through native `inert` and internal hook guards. Card owners retain command dispatch and visuals. `useCardHotkey` clears selection before keyboard commands; clicks use document-owned dismissal.
- Select cards through feature-owned `cardByInteractionTarget`. `useCardInteraction` alone owns semantic card replacement/unmount cleanup; `useDraggableCard` owns native registration. Onboarding retains its isolated provider-level selector.
- `Playground` selects decks directly. `useDroppableTarget` owns `isValid` highlighting and the enablement gate; `PlaygroundDeck` validates live Redux source identity and destination legality in its drop callback before dispatch. Click placement checks `isValid`.
- Keep `PlayerStackDeck` prop-free and available whenever its provider is available, including the empty-stack shortcut.

## Interaction trade-offs and unresolved review decisions

- **Double-click placement remains open.** Two open-stack pile clicks before the server echo can dispatch two identical placement commands. Invalid-pile prevalidation is fixed. The review thread asks whether to clear selection on accepted dispatch or retain focus with an in-flight guard.
- **Activation currently occurs on click/release, not historical mousedown (LIG-144, `8c6e400b`).** This separates selection from dragging but delays non-draggable activation until release. Press-drag-off-release no longer activates cards. A mouse source dragged past 6 px and released over itself does not select because dnd-kit suppresses the follow-up click. The README documents this trade-off; confirmation is requested before changing it.
- **Escape-only deselection needs clarification.** Focused Escape remains a no-op and repeated row hotkeys retain selection, consistent with prior clear-first and custom-cancellation removal feedback. The thread asks whether enabled-only Escape deselection should return without changing native drag cancellation.

These unresolved behaviors are not claimed fixed. No merge is requested.

## Validation of latest review fix (`05b2201`)

- Observed RED/GREEN for consumer-owned drop acceptance while retaining invalid highlighting. Added live Redux source/destination regression cases without waiting for rerender.
- Frontend: **116 tests in 17 files**, typecheck and production build pass.
- Repository `pnpm test:ci`: **250 tests pass**. `pnpm fmt:check`, `pnpm lint:check` and `git diff --check` pass.
- Independent review of the exact incremental diff passed with no security concerns or logic errors before commit.
- Repository typecheck: ten packages pass; Blog reproduces the same three generated `.next/types/validator.ts` TS2344 locale constraint errors at lines 81, 94 and 103 on an independently installed unchanged baseline. No new type errors or Blog changes are included.
- This incremental fix reran unit tests/build, not local browser E2E. Earlier Chromium verification covered native inert disablement. Existing nonblocking Vite/Sentry/build warnings remain.
- GitHub CI restarted on `05b2201`; results are pending. Storybook requires protected-environment approval from the core team, which the worker cannot provide.

Closes #685


apps/ligretto-frontend/src/features/playground/ui/Playground.tsx
```diff
@@ -25,9 +26,11 @@ export const Playground = ({ cardsDecks, onDeckClick, ref, deckRefs }: Playgroun
   return (
     <TableCards ref={ref}>
       {cards.map((card, index) => (
-        <CardPlace key={index} size="large" ref={deckRefs?.[index]} dataTestId={`Playground-Deck-${index}`}>
```
StepanDyubin, line 28: Drop target should be inside CardPlace.

Create PlaygroundDeck component that encapsulates logic inside by getting just cardDeck prop.

apps/ligretto-frontend/src/features/playground/ui/Playground.tsx
StepanDyubin: Why we need that?

apps/ligretto-frontend/e2e/card-placement-harness.html
```diff
@@ -0,0 +1,12 @@
+<!doctype html>
```
StepanDyubin: Remove e2e tests for dnd for now

apps/ligretto-frontend/src/entities/card/ui/Card/Card.spec.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+import { cleanup, fireEvent, render, screen } from '@testing-library/react'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+
+import { CardColors } from '@memebattle/ligretto-shared'
+import { Card } from './Card'
+
+afterEach(cleanup)
+
+describe('Card activation', () => {
```
StepanDyubin: useless test, remove

apps/ligretto-frontend/src/features/cardPlacement/ui/CardPlacementProvider.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const placeCard = useCallback(
+    (target: CardPlacementTarget, playgroundDeckIndex: number) => {
+      if (!enabled) {
+        return false
+      }
+      dispatch(getCardPlacementAction(target, gameId, playgroundDeckIndex))
+      return true
+    },
+    [dispatch, enabled, gameId],
+  )
```
StepanDyubin: That logic should be inside droppable component

apps/ligretto-frontend/src/features/cardPlacement/ui/CardPlacementProvider.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  useEffect(() => {
+    if (!enabled) {
+      setActiveDrag(undefined)
+    }
+  }, [enabled])
+
+  const registerSource = useCallback((source: CardDragData) => {
```
StepanDyubin: Why we need that logic?

apps/ligretto-frontend/src/features/cardPlacement/ui/CardPlacementProvider.tsx
```diff
@@ -0,0 +1,106 @@
...
+    if (dragged && sources.current.has(dragged.id)) {
+      setActiveDrag(dragged)
+    }
+  }
+
+  const handleDragEnd = ({ active, over }: DragEndEvent) => {
+    const dragged = active.data.current as CardDragData | undefined
+    const drop = over?.data.current as PlaygroundDropData | undefined
+    const currentSource = dragged && sources.current.get(dragged.id)
+    if (dragged && currentSource === dragged && drop && canPlaceCardOnDeck(dragged.card, drop.deck)) {
```
StepanDyubin: so we have to call callback of droppable element here, passwed through `data`

apps/ligretto-frontend/src/features/cardPlacement/ui/DraggableCard.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+interface DraggableCardProps extends PropsWithChildren {
+  target: CardPlacementTarget
+  card: Card
+  disabled?: boolean
+}
+
+const getTargetKey = (target: CardPlacementTarget) => (target.type === 'row' ? `row.${target.index}` : target.type)
+
+export const DraggableCard = ({ target, card, disabled = false, children }: DraggableCardProps) => {
```
StepanDyubin: Do not create a component wrapper. Just useDraggableCard hook that keep logic inside.

The same for droppable

StepanDyubin (changes requested):

apps/ligretto-frontend/src/features/playground/ui/Playground.tsx
memebattle-dev: Updated in b488870. Added `PlaygroundDeck`; it receives the complete deck, extracts the visible card internally, and applies the droppable surface inside `CardPlace` with explicit responsive dimensions.

apps/ligretto-frontend/src/features/playground/ui/Playground.tsx
memebattle-dev: Removed the parent-level `useMemo`/top-card projection in b488870. `Playground` now maps decks directly and `PlaygroundDeck` owns top-card extraction.

apps/ligretto-frontend/e2e/card-placement-harness.html
memebattle-dev: Removed the DnD E2E harness, spec, and related Playwright configuration in b488870.

apps/ligretto-frontend/src/entities/card/ui/Card/Card.spec.tsx
memebattle-dev: Removed `Card.spec.tsx` in b488870.

apps/ligretto-frontend/src/features/cardPlacement/ui/CardPlacementProvider.tsx
memebattle-dev: Moved placement and validation out of `CardPlacementProvider` in b488870. `useDroppableCard` validates the current deck and invokes its callback; `PlaygroundContainer` owns Redux command routing.

apps/ligretto-frontend/src/features/cardPlacement/ui/CardPlacementProvider.tsx
memebattle-dev: Removed the source registry and its lifecycle logic from the provider in b488870.

apps/ligretto-frontend/src/features/cardPlacement/ui/CardPlacementProvider.tsx
memebattle-dev: Implemented in b488870: `handleDragEnd` now reads the droppable data and calls its `onDrop(dragged)` callback.

apps/ligretto-frontend/src/features/cardPlacement/ui/DraggableCard.tsx
memebattle-dev: Replaced both wrapper components in b488870. Card owners now call `useDraggableCard` and spread its result directly onto `Card`; `PlaygroundDeck` calls `useDroppableCard`.

apps/ligretto-frontend/src/entities/card/ui/Card/Card.tsx
```diff
@@ -1,14 +1,14 @@
...
+import { ButtonBase, type ButtonBaseProps } from '@mui/material'
 import { Typography } from '@memebattle/ui'
 import { styled } from '@mui/material/styles'
 import type { CardPlaceSize } from '../CardPlace'
 import { CardBackFace } from './CardBackFace'
 
 type CardSize = 'small' | 'medium' | 'large'
 
-interface CardProps {
+interface CardProps extends Omit<ButtonBaseProps, 'color' | 'onClick' | 'size' | 'value'> {
```
StepanDyubin: Remove extends. If you need some prop add them without inheritance

apps/ligretto-frontend/src/features/cardPlacement/ui/CardPlacementContext.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const useCardPlacement = () => {
+  const context = useContext(CardPlacementContext)
+  if (!context) {
+    throw new Error('Card placement hooks must be used within CardPlacementProvider')
+  }
+  return context
+}
+
+export const useOptionalCardPlacement = () => useContext(CardPlacementContext)
```
StepanDyubin: Remove it, always use the one that garanties the context

apps/ligretto-frontend/src/features/cardPlacement/ui/useDroppableCard.ts
```diff
@@ -0,0 +1,45 @@
...
+  const isValid = !!activeDrag && canPlaceCardOnDeck(activeDrag.card, cardDeck)
+  const boxShadow = activeDrag
+    ? isValid
+      ? isOver
+        ? '0 0 0 0.3rem #6ee7a0'
+        : '0 0 0 0.2rem rgba(110, 231, 160, 0.7)'
+      : isOver
+        ? '0 0 0 0.3rem #f87171'
+        : undefined
+    : undefined
```
StepanDyubin: UI component that uses the hook should be responsible for styling. Make hook return only properties like (`isOver`, `isValid` etc)

apps/ligretto-frontend/src/features/cardPlacement/model/getCardPlacementAction.ts
```diff
@@ -0,0 +1,8 @@
+import { putCardAction, putCardFromStackOpenDeck } from '@memebattle/ligretto-shared'
+
+import type { CardPlacementTarget } from './types'
+
+export const getCardPlacementAction = (target: CardPlacementTarget, gameId: string, playgroundDeckIndex: number) =>
```
StepanDyubin: Remove that function. Inline code in components

apps/ligretto-frontend/src/features/cardPlacement/ui/useDraggableCard.ts
```diff
@@ -0,0 +1,24 @@
...
+  const data = useMemo<CardDragData>(() => ({ target, card }), [card, target])
+  const { attributes, isDragging, listeners, setNodeRef } = useDraggable({ id, data, disabled: disabled || !enabled })
+
+  return {
+    ...attributes,
+    ...listeners,
+    ref: setNodeRef,
+    'data-card-drag-source': true,
+    'data-card-drag-id': id,
+    style: { opacity: isDragging ? 0.35 : 1, touchAction: 'none' },
```
StepanDyubin: Do not return styles, just properties to allow a component do styling itself

apps/ligretto-gameplay-backend/src/entities/playground/playground.service.ts
```diff
@@ -22,13 +14,13 @@ export class PlaygroundService {
 
   findAvailableDeckIndex(gameId: UUID, card: Card) {
```
StepanDyubin: We dont need that since auto placement is deleted

apps/ligretto-frontend/src/features/cardPlacement/ui/useDroppableCard.ts
```diff
@@ -0,0 +1,45 @@
+import { useMemo } from 'react'
+import { useDroppable } from '@dnd-kit/core'
+import { canPlaceCardOnDeck, type CardsDeck } from '@memebattle/ligretto-shared'
+
+import type { CardDragData, PlaygroundDropData } from '../model/types'
+import { useOptionalCardPlacement } from './CardPlacementContext'
+
+export const useDroppableCard = (id: string, cardDeck: CardsDeck | null | undefined, onDrop: (dragged: CardDragData) => void) => {
+  const placement = useOptionalCardPlacement()
```
StepanDyubin: Remove optional, it should always be inside context

StepanDyubin (commented):
@memebattle-dev fix

apps/ligretto-frontend/src/entities/card/ui/Card/Card.tsx
memebattle-dev: Fixed in fbfab3c. `CardProps` no longer extends `ButtonBaseProps`; it declares only the specific ref, style, and pointer/key handler props needed by DnD.

apps/ligretto-frontend/src/features/cardPlacement/ui/CardPlacementContext.ts
memebattle-dev: Removed the optional context accessor in fbfab3c. Both DnD hooks now use the guaranteed context hook, and onboarding/tests are explicitly wrapped in `CardPlacementProvider`.

apps/ligretto-frontend/src/features/cardPlacement/ui/useDroppableCard.ts
memebattle-dev: Updated in fbfab3c. `useDroppableCard` now returns only `isOver`, `isValid`, and `setNodeRef`; `PlaygroundDeck` owns dimensions, visual styles, and data attributes.

apps/ligretto-frontend/src/features/cardPlacement/model/getCardPlacementAction.ts
memebattle-dev: Removed `getCardPlacementAction` and its spec in fbfab3c. `PlaygroundContainer` now constructs the two existing placement actions inline.

apps/ligretto-frontend/src/features/cardPlacement/ui/useDraggableCard.ts
memebattle-dev: Updated in fbfab3c. `useDraggableCard` returns only ID, dragging state, listeners, and node ref; row/open-stack components own opacity, touch styling, and data attributes.

apps/ligretto-gameplay-backend/src/entities/playground/playground.service.ts
memebattle-dev: Removed `findAvailableDeckIndex` and the optional fallback in fbfab3c. Automatic placement is now fully removed: destination indices are required, value-1 cards use focus/explicit placement, and obsolete tap actions/listeners were deleted.

apps/ligretto-frontend/src/features/cardPlacement/ui/useDroppableCard.ts
memebattle-dev: Removed optional context usage in fbfab3c. `useDroppableCard` now always requires `CardPlacementProvider`.

memebattle-dev:
Fixed the onboarding E2E regressions in dc0b9f7 and 2ef5127: playground interactions now target the deck drop surface, including empty decks, and the free-play scenario performs the required manual select-then-place sequence. Frontend typecheck and all 76 unit tests pass; repository formatting and lint checks pass. The complete Ligretto CI rerun is green, including all 9 Playwright tests across desktop and mobile projects.

apps/ligretto-frontend/src/features/cardInteraction/ui/useDroppableCard.ts
```diff
@@ -0,0 +1,28 @@
...
+import { canPlaceCardOnDeck, type CardsDeck } from '@memebattle/ligretto-shared'
+
+import type { CardDragData, CardDropData } from '../model/types'
+import { useCardInteractionContext } from './CardInteractionContext'
+
+export const useDroppableCard = (id: string, cardDeck: CardsDeck | null | undefined, onDrop: (dragged: CardDragData) => void) => {
+  const { activeCard, enabled } = useCardInteractionContext()
+  const data = useMemo<CardDropData>(
+    () => ({
+      onDrop: dragged => {
```
StepanDyubin: Let's add also onDragOver
And each consumer of useDroppableCard will be allowed to define logic inside (e.g. isValid)

apps/ligretto-frontend/src/features/player/ui/PlayerCardsStack/PlayerStackDeck.tsx
```diff
@@ -14,9 +14,9 @@ interface PlayerStackDeckProps {
 
 export const PlayerStackDeck = ({ card, enabled, isHidden }: PlayerStackDeckProps) => {
   const dispatch = useDispatch()
-  const { clearFocus } = useCardFocus()
+  const { clearActiveTarget } = useCardInteraction()
   const onStackDeckActivate = () => {
-    clearFocus()
+    clearActiveTarget()
```
StepanDyubin: If it's a click, so target should clear by listener on document.
If it's hotkey, lets move clear logic inside `useCardHotkey`, move it to cardInteraction feature.

So here wi won't call clear. The same for other card components

apps/ligretto-frontend/src/features/cardInteraction/ui/useDroppableCard.ts
memebattle-dev: Addressed in b8e50f8. `useDroppableCard` now routes consumer-owned `onDragOver` and `onDrop` callbacks, while `PlaygroundDeck` owns placement validation and validity styling.

apps/ligretto-frontend/src/features/player/ui/PlayerCardsStack/PlayerStackDeck.tsx
memebattle-dev: Addressed in b8e50f8. `useCardHotkey` now lives in `cardInteraction` and owns clearing for command hotkeys; card click clearing remains document-driven, so card components no longer call `clearActiveTarget` directly.

apps/ligretto-frontend/src/features/player/ui/PlayerRowCardsContainer/PlayerRowCardsContainer.tsx
```diff
@@ -15,21 +14,26 @@ interface PlayerRowCardProps {
...
+        {...card}
+        {...listeners}
+        ref={setNodeRef}
+        data-card-drag-source
+        data-card-drag-id={dragId}
+        data-card-interaction-element
+        isDarkened={isDimmed}
+        isSelected={isActive}
+        onClick={onCardActivate}
+        style={{ opacity: isDragging ? 0.35 : 1, touchAction: 'none' }}
```
StepanDyubin: Dnd should feel like we drag the card, so do not show faded card in original place while dragging

apps/ligretto-frontend/src/features/cardInteraction/ui/useDroppableCard.ts
```diff
@@ -0,0 +1,17 @@
+import { useMemo } from 'react'
+import { useDroppable } from '@dnd-kit/core'
+
+import type { CardDragData, CardDropData } from '../model/types'
+import { useCardInteractionContext } from './CardInteractionContext'
+
+export const useDroppableCard = (id: string, onDrop: (dragged: CardDragData) => void, onDragOver?: (dragged: CardDragData) => void) => {
```
StepanDyubin: Remove onDragOver since isOver exposed be useDroppable

StepanDyubin: Rename to useDroppableTarget.

pass target instead of id - compose id inside the hook

apps/ligretto-frontend/src/features/cardInteraction/ui/useCardHotkey.ts
```diff
@@ -0,0 +1,23 @@
+import { useHotkeys } from 'react-hotkeys-hook'
+
+import type { Hotkey } from '#ducks/game'
+import { useCardInteraction } from './useCardInteraction'
+
+interface CardHotkeyOptions {
+  clearActive?: boolean
+}
+
+export const useCardHotkey = (hotkey: Hotkey | undefined, onActivate: () => void, { clearActive = true }: CardHotkeyOptions = {}) => {
```
StepanDyubin: Any hotkey interaction should clear active target first. So remove options

apps/ligretto-frontend/src/features/player/ui/PlayerRowCardsContainer/PlayerRowCardsContainer.tsx
memebattle-dev: Fixed in c3dce9d. Row and open-stack source cards now use opacity `0` while dragging, so only the drag overlay remains visible.

apps/ligretto-frontend/src/features/cardInteraction/ui/useDroppableCard.ts
memebattle-dev: Removed `onDragOver` from drop data, the provider, and hook API in c3dce9d. Consumers use the exposed `isOver` state.

memebattle-dev: Renamed it to `useDroppableTarget` in c3dce9d. It now accepts a semantic `CardDropTarget` and composes/returns the DnD ID internally.

apps/ligretto-frontend/src/features/cardInteraction/ui/useCardHotkey.ts
memebattle-dev: Removed the options in c3dce9d. Every `useCardHotkey` activation now clears `activeTarget` before invoking the owner callback; consumers and tests were updated.

apps/ligretto-frontend/src/features/cardInteraction/ui/useDroppableTarget.ts
```diff
@@ -0,0 +1,21 @@
+import { useMemo } from 'react'
+import { useDroppable } from '@dnd-kit/core'
+
+import type { CardDragData, CardDropData, CardDropTarget } from '../model/types'
+import { useCardInteractionContext } from './CardInteractionContext'
+
+const getDropTargetKey = (target: CardDropTarget) => `${target.type}.${target.index}`
```
StepanDyubin: use iteraction target also here, extend the type.

reuse getIteractionTarkeyKey across feature

apps/ligretto-frontend/src/features/cardInteraction/ui/CardInteractionProvider.tsx
```diff
@@ -0,0 +1,123 @@
...
+import type { CardDragData, CardDropData, CardInteractionTarget } from '../model/types'
+import { CardInteractionContext, isSameCardInteractionTarget } from './CardInteractionContext'
+
+interface CardInteractionProviderProps extends PropsWithChildren {
+  enabled: boolean
+}
+
+interface InteractionState {
+  activeTarget?: CardInteractionTarget
+  activeCard?: Card
```
StepanDyubin: Do not store activeCard

get card value by using selector where it is nessesary

apps/ligretto-frontend/src/features/playground/ui/PlaygroundDeck/PlaygroundDeck.tsx
```diff
@@ -0,0 +1,72 @@
...
+    width: mobileWidthBySize.large,
+    height: mobileHeightBySize.large,
+  },
+}))
+
+interface PlaygroundDeckProps {
+  cardDeck: CardsDeck | null | undefined
+  deckIndex: number
+  onClick: () => void
+  onDrop?: (dragged: CardDragData) => void
```
StepanDyubin: Do not pass handlers through.
just use redux inside the component. keep only cardDeck and deckIndex

apps/ligretto-frontend/src/features/cardInteraction/ui/useDroppableTarget.ts
memebattle-dev: Addressed in ba5cfaa. `CardInteractionTarget` now includes playground targets, and both draggable and droppable hooks reuse `getInteractionTargetKey`; target lifecycle cleanup uses the same key.

apps/ligretto-frontend/src/features/cardInteraction/ui/CardInteractionProvider.tsx
memebattle-dev: Addressed in ba5cfaa. The interaction context stores only `activeTarget`; the drag overlay and drop validity now select the current card from Redux where needed.

apps/ligretto-frontend/src/features/playground/ui/PlaygroundDeck/PlaygroundDeck.tsx
memebattle-dev: Addressed in ba5cfaa. `PlaygroundDeck` now accepts only `cardDeck` and `deckIndex` and owns Redux click/drop placement. Onboarding uses its own tested playground view to preserve scripted actions and refs.

apps/ligretto-frontend/src/features/cardInteraction/ui/useDroppableTarget.ts
memebattle-dev: Implemented in ba5cfaa. Playground is now part of `CardInteractionTarget`; draggable and droppable hooks reuse the shared `getInteractionTargetKey`.

apps/ligretto-frontend/src/features/cardInteraction/ui/CardInteractionProvider.tsx
memebattle-dev: Implemented in ba5cfaa. `activeCard` was removed from interaction state/context; the provider overlay and playground decks derive current card data from Redux selectors.

apps/ligretto-frontend/src/features/playground/ui/PlaygroundDeck/PlaygroundDeck.tsx
memebattle-dev: Implemented in ba5cfaa. `PlaygroundDeck` now accepts only `cardDeck` and `deckIndex`, and owns Redux selection/dispatch, click placement, drop placement, and validity logic internally.

memebattle-dev:
Implemented the component-owned interaction refinement in 4743ab15782684a896d4d8b6fd7c499f6889115b.

- `cardInteraction` owns transient target state, input gating, sensors, cancellation and drop delivery through hooks. The provider no longer selects Redux data, renders card UI or invokes placement callbacks.
- Components retain target rules, Redux commands, card data and styling. Gameplay drag preview now lives in `PlayerCardDragOverlay` and consumes `useCardDragTarget`.
- Hook-level guards keep disabled inputs, active gestures and stale source replacement from dispatching commands; onboarding remains isolated from gameplay commands.

Validation: frontend typecheck and all 97 unit tests passed; build, formatter, lint and independent review passed. Local onboarding E2E could not reach the tutorial because the core backend at 127.0.0.1:3333 was unavailable (`/api/auth/me`: connection refused). CI is running on this commit.

apps/ligretto-frontend/src/features/player/ui/PlayerCardsStack/PlayerCardsStack.tsx
```diff
@@ -27,17 +28,19 @@ const ReshuffleHint = styled('div')(({ theme }) => ({
...
 
-  const isStackDeckEnabled = stackDeckCards.length > 0 || !!stackOpenDeckCard
+  const isStackDeckEnabled = inputEnabled && (stackDeckCards.length > 0 || !!stackOpenDeckCard)
 
   return (
     <CardsRow>
       <CardPlace>
         {stackOpenDeckCard && (
-          <CardHotkeyBadge hotkey={Hotkey.x}>
+          <CardHotkeyBadge hotkey={inputEnabled ? Hotkey.x : undefined}>
```
StepanDyubin: hotkey should always be available, and cancel drag if it is in process

apps/ligretto-frontend/src/features/cardInteraction/ui/useCardInputEnabled.ts
```diff
@@ -0,0 +1,10 @@
+import { useDndContext } from '@dnd-kit/core'
+import { useCardInteractionContext } from './CardInteractionContext'
+
+// A native drag owns input until release/cancel; clicks and shortcuts must not
+// replace its selection or resurrect a gesture invalidated by a card update.
+export const useCardInputEnabled = () => {
```
StepanDyubin: So remove that hook

apps/ligretto-frontend/src/features/player/ui/PlayerCardsStack/PlayerStackDeck.tsx
StepanDyubin: Isn't that always hidden? Can we remove that prop, and `card` also?

apps/ligretto-frontend/src/features/cardInteraction/ui/useCardInputEnabled.ts
StepanDyubin: Do not pass `enabled` to no components inside a provider. The provider should encapsulate all logic that disables interactions inside.

apps/ligretto-frontend/src/features/cardInteraction/ui/useDroppableTarget.ts
```diff
@@ -0,0 +1,32 @@
...
+      // dnd-kit retains active data after a source disappears. Only a live,
+      // unchanged source and destination may complete the original gesture.
+      if (
+        enabled &&
+        activeTarget &&
+        dragged?.target &&
+        over?.id === id &&
+        draggableNodes.get(active.id)?.node.current?.isConnected &&
+        droppableContainers.get(id)?.node.current?.isConnected &&
+        isSameCardInteractionTarget(activeTarget, dragged.target)
```
StepanDyubin: Why is it better than it was in previous version (passing handler through `data`)?

apps/ligretto-frontend/src/features/cardInteraction/ui/CardInteractionProvider.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+import type { CardInteractionTarget } from '../model/types'
+import { CardInteractionContext, isSameCardInteractionTarget } from './CardInteractionContext'
+
+interface CardInteractionProviderProps extends PropsWithChildren {
+  enabled: boolean
+}
+
+export const CardInteractionProvider = ({ children, enabled }: CardInteractionProviderProps) => {
+  const [activeTarget, setActiveTarget] = useState<CardInteractionTarget>()
```
StepanDyubin: use reducer here (like a state-machine)
```
type CardInteractionState =
  | { mode: 'idle' }
  | {
      mode: 'focused'
      target: CardTarget
    }
  | {
      mode: 'dragging'
      target: CardTarget
      card: Card // not sure if it is nessesary
    }
```

apps/ligretto-frontend/src/features/cardInteraction/ui/useCardAction.ts
```diff
@@ -0,0 +1,16 @@
+import { useCallback } from 'react'
+import { useCardInteractionContext } from './CardInteractionContext'
+import { useCardInputEnabled } from './useCardInputEnabled'
+
+export const useCardAction = (onActivate: () => void, available = true) => {
```
StepanDyubin: remove that

apps/ligretto-frontend/src/features/player/ui/PlayerCardsStack/PlayerCardsStack.tsx
memebattle-dev: Implemented in 7048d9b. Hotkeys stay available during both active and pending mouse/touch gestures. The native sensor is cancelled before the component-owned command runs; regression tests cover new focus surviving the abandoned release/click and a fresh subsequent drag.

apps/ligretto-frontend/src/features/cardInteraction/ui/useCardInputEnabled.ts
memebattle-dev: Implemented in 7048d9b, including the provider-encapsulation follow-up. Removed useCardInputEnabled and consumer-level global enabled propagation. The provider now blocks disabled pointer/click input and hotkey commands and cancels pending/active native gestures; components retain only their own game-specific availability rules.

apps/ligretto-frontend/src/features/player/ui/PlayerCardsStack/PlayerStackDeck.tsx
memebattle-dev: Implemented in 7048d9b. Interaction state is now a useReducer discriminated union with idle, focused and dragging states. Selection and drop validation use that state; imperative refs are limited to native sensor cancellation and abandoned-release cleanup.

apps/ligretto-frontend/src/features/cardInteraction/ui/useDroppableTarget.ts
memebattle-dev: Removed in 7048d9b. Pointer commands are direct component-owned callbacks again; useCardHotkey handles keyboard input and the provider handles cancellation/disablement. There is no useCardAction export or hook.

apps/ligretto-frontend/src/features/cardInteraction/ui/CardInteractionProvider.tsx
memebattle-dev: Implemented in 7048d9b. PlayerStackDeck now receives no props and selects its own stack data. Populated closed-stack cards remain face-down; the empty-stack placeholder and reshuffle availability are preserved and covered by owner tests.

apps/ligretto-frontend/src/features/cardInteraction/ui/useCardAction.ts
memebattle-dev: Restored in 7048d9b. useDroppableTarget puts onDrop in the droppable data again instead of subscribing each target with useDndMonitor. The provider routes a valid current drag to that callback; placement rules and commands remain in the owning component.

apps/ligretto-frontend/src/features/player/ui/PlayerCardsStack/PlayerCardsStack.tsx
memebattle-dev: Following the latest follow-up, removed all custom drag cancellation in e104ef0. Standard MouseSensor/TouchSensor replace the adapters; forced Escape interception, sensor cancellation refs and release-click suppression are gone. Hotkeys remain available and no longer cancel or invalidate an ongoing drag: the original source can still drop normally. Only standard dnd-kit lifecycle synchronization remains. Frontend typecheck, 98 tests and build pass; the independent review found no blockers.

apps/ligretto-frontend/src/features/player/ui/PlayerCardsStack/PlayerStackDeck.tsx
```diff
@@ -1,26 +1,21 @@
...
-  isHidden: boolean
-}
-
-export const PlayerStackDeck = ({ card, enabled, isHidden }: PlayerStackDeckProps) => {
+export const PlayerStackDeck = () => {
   const dispatch = useDispatch()
-  const { clearFocus } = useCardFocus()
+  const { stackDeckCards, stackOpenDeckCard } = useSelector(playerCardsStackSelector)
+  const hasCards = !!stackDeckCards?.length
+  const available = hasCards || !!stackOpenDeckCard
```
StepanDyubin: Remove `available`. It should be always available if provider is avalable

apps/ligretto-frontend/src/features/cardInteraction/ui/CardInteractionProvider.tsx
```diff
@@ -0,0 +1,147 @@
...
+      dispatch({ type: 'dragTerminal', sourceId: active.id })
+    },
+  })
+  return null
+}
+
+export const CardInteractionProvider = ({ children, enabled }: CardInteractionProviderProps) => {
+  const [state, dispatch] = useReducer(reducer, { mode: 'idle' })
+  const enabledRef = useRef(enabled)
+  enabledRef.current = enabled
```
StepanDyubin: remove that ref. and runCommand

apps/ligretto-frontend/src/features/cardInteraction/ui/useCardHotkey.ts
```diff
@@ -1,14 +1,19 @@
 import { useHotkeys } from 'react-hotkeys-hook'
+import { useCallback } from 'react'
 
 import type { Hotkey } from '#ducks/game'
+import { useCardInteractionContext } from './CardInteractionContext'
 
 export const useCardHotkey = (hotkey: Hotkey | undefined, onActivate: () => void) => {
+  const { runCommand } = useCardInteractionContext()
+  const activate = useCallback(() => runCommand(onActivate), [onActivate, runCommand])
```
StepanDyubin: useCardHotkey should not return activate

apps/ligretto-frontend/src/features/cardInteraction/ui/useDraggableCard.ts
```diff
@@ -0,0 +1,29 @@
+import { useEffect, useMemo } from 'react'
+import { useDraggable } from '@dnd-kit/core'
+import type { Card } from '@memebattle/ligretto-shared'
+
+import type { CardDragData, CardDragTarget } from '../model/types'
+import { getInteractionTargetKey, useCardInteractionContext } from './CardInteractionContext'
+
+export const useDraggableCard = (target: CardDragTarget, card: Card, disabled = false) => {
```
StepanDyubin: if component is rendered and uses that hook, it should be enable. Remove `disabled`

apps/ligretto-frontend/src/features/cardInteraction/ui/CardInteractionProvider.tsx
```diff
@@ -0,0 +1,147 @@
...
+  const value = useMemo(
+    () => ({ activeTarget, clearActiveTarget, toggleActiveTarget, runCommand }),
+    [activeTarget, clearActiveTarget, runCommand, toggleActiveTarget],
+  )
+  const captureInput = (event: React.SyntheticEvent) => {
+    if (!enabled) {
+      event.preventDefault()
+      event.stopPropagation()
+    }
+  }
```
StepanDyubin: remove that

apps/ligretto-frontend/src/features/playground/ui/PlaygroundContainer.tsx
```diff
@@ -1,38 +1,11 @@
...
-import { gameIdSelector, playgroundDecksSelector } from '#ducks/game'
-import { useCardFocus } from '#features/cardFocus'
-
-const PlaygroundContainerSelector = createSelector([playgroundDecksSelector, gameIdSelector], (playgroundDecks, gameId) => ({
-  playgroundDecks,
-  gameId,
-}))
+import { playgroundDecksSelector } from '#ducks/game'
 
 export const PlaygroundContainer = () => {
```
StepanDyubin: Remove Container, just use selectors inside Playground component

apps/ligretto-frontend/src/features/playground/ui/PlaygroundDeck/PlaygroundDeck.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const handleDrop = useCallback(
+    (dragged: CardDragData) => {
+      if (canPlaceCardOnDeck(dragged.card, cardDeck)) {
+        placeCard(dragged.target)
+      }
+    },
+    [cardDeck, placeCard],
+  )
+  const { id: dropId, isOver, setNodeRef } = useDroppableTarget({ type: 'playground', index: deckIndex }, handleDrop)
+  const isValid = !!activeCard && canPlaceCardOnDeck(activeCard, cardDeck)
```
StepanDyubin: return isValid from useDroppableTarget.
move selectors call and check inside the hook

apps/ligretto-frontend/src/features/cardInteraction/ui/useCardInteraction.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+export function useCardInteraction(
+  target: CardInteractionTarget,
+  deps: DependencyList,
+): {
+  isActive: boolean
+  isDimmed: boolean
+  toggleActiveTarget: () => void
+}
+export function useCardInteraction(target?: CardInteractionTarget, deps: DependencyList = []) {
```
StepanDyubin: Do not pass deps.
call selector inside and use values to clear effect

StepanDyubin: create cardByInteractionTarget selector inside the feature

apps/ligretto-frontend/src/features/player/ui/PlayerCardsStack/PlayerStackDeck.tsx
memebattle-dev: Addressed in a1554e3. Removed `available` and the conditional stack handlers. The empty-stack Space shortcut remains available whenever the provider is enabled; owner tests cover the shortcut and exact dispatch.

Validation: 105 frontend tests, frontend typecheck/build, repository unit tests (234), formatting and lint pass; independent review found no blockers.

apps/ligretto-frontend/src/features/cardInteraction/ui/CardInteractionProvider.tsx
memebattle-dev: Addressed in a1554e3. Removed `enabledRef` and `runCommand`. Hotkey clearing lives in `useCardHotkey`; provider state and internal hooks retain the enablement gate without consumer flags or replacement command machinery.

Validation: 105 frontend tests, frontend typecheck/build, repository unit tests (234), formatting and lint pass; independent review found no blockers.

apps/ligretto-frontend/src/features/cardInteraction/ui/useCardHotkey.ts
memebattle-dev: Addressed in a1554e3. `useCardHotkey` now returns nothing. It clears selection before the keyboard callback, while clicks invoke owner callbacks directly and use document-driven dismissal. The public API and owner dispatch paths are covered by tests.

Validation: 105 frontend tests, frontend typecheck/build, repository unit tests (234), formatting and lint pass; independent review found no blockers.

apps/ligretto-frontend/src/features/cardInteraction/ui/useDraggableCard.ts
memebattle-dev: Addressed in a1554e3. Removed the `disabled` parameter and drag-data property. Mounted sources are available under an enabled provider; the hook reads global provider enablement internally. Native mouse/touch regressions cover disablement and re-enablement.

Validation: 105 frontend tests, frontend typecheck/build, repository unit tests (234), formatting and lint pass; independent review found no blockers.

apps/ligretto-frontend/src/features/cardInteraction/ui/CardInteractionProvider.tsx
memebattle-dev: Addressed in a1554e3. Removed `captureInput` and all capture handlers. The provider uses native `inert` for pointer/focus disablement, with internal hotkey/drag guards. A real Chromium check against `PlayerStackDeck` verifies enabled click/Space dispatch exactly once and disabled pointer/focus/Space dispatch nothing, including after re-disabling.

Validation: 105 frontend tests, frontend typecheck/build, repository unit tests (234), formatting and lint pass; independent review found no blockers.

apps/ligretto-frontend/src/features/playground/ui/PlaygroundContainer.tsx
memebattle-dev: Addressed in a1554e3. Deleted `PlaygroundContainer`; `Playground` selects its decks directly. Updated its export and GameContainer consumer, with rendered tests for row/open-stack placement and exact payloads.

Validation: 105 frontend tests, frontend typecheck/build, repository unit tests (234), formatting and lint pass; independent review found no blockers.

apps/ligretto-frontend/src/features/playground/ui/PlaygroundDeck/PlaygroundDeck.tsx
memebattle-dev: Addressed in a1554e3. `useDroppableTarget` now owns the card/deck selectors and returns `isValid`. It also validates the live source and destination before delivering the callback; `PlaygroundDeck` retains only command dispatch and visuals. Tests cover changing source/deck data and drag identity after hotkey focus changes.

Validation: 105 frontend tests, frontend typecheck/build, repository unit tests (234), formatting and lint pass; independent review found no blockers.

apps/ligretto-frontend/src/features/cardInteraction/ui/useCardInteraction.ts
memebattle-dev: Addressed in a1554e3. Removed caller-provided dependencies and added the feature-owned `cardByInteractionTarget` selector. The hook clears on selected card color/value changes and unmount, not object reallocation. The overlay reuses the selector; onboarding supplies its isolated tutorial selector at the provider boundary. Selection replacement and tutorial isolation regressions pass.

Validation: 105 frontend tests, frontend typecheck/build, repository unit tests (234), formatting and lint pass; independent review found no blockers.

apps/ligretto-frontend/src/features/cardInteraction/ui/CardInteractionProvider.tsx
```diff
@@ -0,0 +1,130 @@
...
+    },
+  })
+  return null
+}
+
+export const CardInteractionProvider = ({ children, enabled, cardByTargetSelector = cardByInteractionTarget }: CardInteractionProviderProps) => {
+  const [state, dispatch] = useReducer(reducer, { mode: 'idle' })
+  const sensors = useSensors(
+    useSensor(MouseSensor, { activationConstraint: { distance: 6 } }),
+    useSensor(TouchSensor, { activationConstraint: { delay: 150, tolerance: 8 } }),
```
Themezv: **Bug: a touch held for 150 ms without moving becomes a drag that ends nowhere, and the tap is lost.**

With a `delay` constraint dnd-kit's `AbstractPointerSensor.attach()` schedules `handleStart` after 150 ms regardless of movement (the `touchDrag` helper in `dnd.spec.tsx` relies on exactly this). So on a phone a deliberate tap of ~150-300 ms:

1. activates the drag: `onDragStart` → `dragStart` replaces any `focused` state with `dragging`;
2. the finger lifts at the same spot, `pointerWithin` gives `over === null`, no `onDrop`, `dragTerminal` → `idle`;
3. dnd-kit installs a capture-phase document `click` listener with `stopPropagation` on activation and keeps it for 50 ms after release, so the card's `onClick` toggle never runs.

Net effect: the card is not selected, and any previous selection is silently cleared. 150 ms is well inside a normal tap; dnd-kit's own docs suggest ~250 ms for touch.

**Proposals, in order of preference:**

- **Use a distance constraint instead of a delay:** `useSensor(TouchSensor, { activationConstraint: { distance: 8 } })`. A still finger then never activates a drag, the click goes through normally and the click suppression never kicks in. There is no scroll conflict because the cards already set `touchAction: 'none'`. This is a one-line change.
- **If the delay must stay:** treat a drag that ends with `over === null` and a displacement below `tolerance` as a tap. `onDragEnd` receives `delta`; dispatch `toggle` instead of `dragTerminal` in that case. This needs the `dragging` state to remember the previous `State` so a tap on an already-selected card deselects it, and it keeps the 150 ms latency before a real drag starts.
- At minimum, raise the delay to ~250 ms so fewer taps fall into the window.

I would also add a unit case: touch start, wait past the delay, touch end without moving → `activeTarget` toggles.

apps/ligretto-frontend/src/features/playground/ui/PlaygroundDeck/PlaygroundDeck.tsx
```diff
@@ -0,0 +1,84 @@
...
+  return (
+    <CardPlace size="large" dataTestId={`Playground-Deck-${deckIndex}`}>
+      <DropSurface
+        ref={setNodeRef}
+        data-card-drop-target={dropId}
+        data-drop-valid={isValid || undefined}
+        data-drop-over={isOver || undefined}
+        onClick={() => {
+          if (activeTarget?.type === 'row' || activeTarget?.type === 'open-stack') {
+            placeCard(activeTarget)
```
Themezv, line 75: **Please verify before fixing: the click path neither validates nor clears, so a double-tap can place a second, unchosen card.**

The drop path now prevalidates in `useDroppableTarget.onDrop`, but this click handler dispatches `placeCard(activeTarget)` unconditionally, and the selection survives the click because `TableCards` carries `data-card-interaction-element`; it is only cleared when the card identity changes after the server echo (there is no optimistic reducer).

Scenario: open-stack card selected, double-tap on a pile → two `putCardFromStackOpenDeck` before `updateGameAction` arrives. Backend `playerPutFromStackOpenDeck` reads `last(deck.cards)` and `pop()`s, so the second request evaluates the *next* open-stack card and places it whenever it happens to be the same colour and value+1. Row cards are safe only because `removeCard` nulls the slot. Every misclick on an invalid pile is also a round trip plus a room-wide broadcast for a no-op.

The README says click selection is intentionally kept until the server confirms, and the base `PlaygroundContainer` had the same shape, so this is carried over rather than new. Suggested minimum: run the same `canPlaceCardOnDeck` check the drop path uses (the `isValid` returned by `useDroppableTarget` is already computed here), and either clear the selection on dispatch or ignore a second click while a put for that target is in flight.

apps/ligretto-frontend/src/entities/card/ui/Card/Card.tsx
```diff
@@ -153,7 +159,7 @@ export const Card: React.FC<CardProps> = ({
     isHighlighted={isHighlighted}
     disableRipple={isDisabled}
     size={size}
-    onMouseDown={onClick}
+    onClick={onClick}
```
Themezv, line 162: **Please verify before fixing: `onMouseDown` → `onClick` is an unacknowledged trade-off, and with the 6 px MouseSensor constraint a jittery press is swallowed.**

`onMouseDown={onClick}` was introduced deliberately in 8c6e400b (LIG-144, "put card on mousedown instead of click") as a speed-game affordance. This PR flips every `Card` to `onClick` without mentioning it in the description or README.

For draggable cards the combination with `activationConstraint: { distance: 6 }` has a concrete failure: mousedown on a row card, drift ≥ 6 px, mouseup over the card itself → dnd-kit activates, `over === null`, `onDragEnd` only dispatches `dragTerminal`, and the 50 ms capture-phase click suppression prevents `onCardActivate`. The card is neither selected nor placed; on the base branch it toggled on mousedown regardless of drift. For non-draggable cards (stack deck, ligretto pack, onboarding) activation now waits for mouseup and press-drag-off-release no longer fires.

If the change is intended, please say so in the PR body and the README. If not, options are a larger distance, or activating on `onMouseDown` for non-draggable cards only.

apps/ligretto-frontend/src/features/cardInteraction/ui/useCardHotkey.ts
```diff
@@ -1,14 +1,17 @@
...
 import type { Hotkey } from '#ducks/game'
+import { useCardInteractionContext } from './CardInteractionContext'
 
 export const useCardHotkey = (hotkey: Hotkey | undefined, onActivate: () => void) => {
+  const { enabled, clearActiveTarget } = useCardInteractionContext()
   useHotkeys(
     hotkey ?? '',
     event => {
       event.preventDefault()
+      clearActiveTarget()
```
Themezv, line 12: **Please verify before fixing: there is no longer any keyboard way to deselect a card.**

The base `CardFocusProvider` bound Escape to clear focus, and a repeated row hotkey toggled the selection off. Now the Escape handler is gone (`Hotkey.escape` in `ducks/game/utils.ts` is dead) and every hotkey does `clearActiveTarget()` then `onActivate()`; after the clear the reducer is `idle`, so `toggle` always ends in `focused`. Pressing `q` twice keeps row 0 selected (asserted in `cardOwnedHotkeys.spec.tsx`), and `index.spec.tsx` asserts Escape is a no-op.

So after `q` the board dims and the only side-effect-free exit is a mouse click outside a card element; Space and L clear only by dispatching a game command. Issue #685 asks to preserve keyboard accessibility.

Suggested: either restore an Escape → `clearActiveTarget()` binding in the provider (gated on `enabled`), or let a repeated hotkey on the already-active target toggle it off by skipping the `clearActiveTarget()` when `onActivate` is a toggle of the same target.

apps/ligretto-frontend/src/features/cardInteraction/ui/useDraggableCard.ts
```diff
@@ -0,0 +1,29 @@
...
+import type { CardDragData, CardDragTarget } from '../model/types'
+import { getInteractionTargetKey, useCardInteractionContext } from './CardInteractionContext'
+
+export const useDraggableCard = (target: CardDragTarget, card: Card) => {
+  const { clearActiveTarget, enabled } = useCardInteractionContext()
+  const id = `${getInteractionTargetKey(target)}.${card.color}.${card.value}`
+  const data = useMemo<CardDragData>(() => ({ target, card }), [card, target])
+  const { isDragging, listeners, setNodeRef } = useDraggable({ id, data, disabled: !enabled })
+
+  useEffect(() => {
```
Themezv: **Please verify before fixing: this cleanup effect duplicates the one in `useCardInteraction`.**

Both production callers (`PlayerRowCard`, `PlayerStackOpenCard`) call `useCardInteraction(target)` and `useDraggableCard(target, card)`. `useCardInteraction` keys its effect on `[clearActiveTarget, targetKey, card?.color, card?.value]` and returns `() => clearActiveTarget(target)`; this effect keys on `id = \`${targetKey}.${card.color}.${card.value}\`` (the same identity) and returns the same cleanup. Every slot refill or unmount therefore dispatches two `clear` actions for one target, each behind its own `eslint-disable`, and the next source owner has to remember to double-hook.

Suggested: let one hook own identity and cleanup. Either drop this effect and keep only the `!enabled` branch, or fold `useDraggableCard` into `useCardInteraction` so a card source is one hook.

Themezv (commented):
Review of the card interaction rewrite. The first comment is a confirmed bug with concrete proposals. The other four are findings I would like the author to verify against the current head before fixing, since the branch moved during the review (the highlight desync, the input-swallowing wrapper, the hotkey enablement and the duplicated target-to-card selector were already fixed in a1554e33).

apps/ligretto-frontend/src/features/cardInteraction/ui/useCardHotkey.ts
memebattle-dev: Confirmed: focused Escape is a no-op and repeated row hotkeys retain selection. Earlier feedback required every hotkey to clear first (https://github.com/MemeBattle/monorepo/pull/686#discussion_r3914852719), and the later follow-up removed custom Escape/drag cancellation (https://github.com/MemeBattle/monorepo/pull/686#discussion_r3942109063). Should we restore enabled-only Escape deselection while leaving native drag cancellation unchanged? No keyboard behavior was changed in this follow-up.

apps/ligretto-frontend/src/features/cardInteraction/ui/CardInteractionProvider.tsx
memebattle-dev: Fixed in 3b5d4f427d513118fb1b418c7eb0c207f9083a2f: touch dragging now requires movement beyond 8 px, so a stationary 300 ms touch reaches normal click selection. Added select/deselect regressions and updated the native touch-drag helper. All 110 frontend tests and 239 repository tests pass; frontend typecheck/build, formatting and lint pass.

apps/ligretto-frontend/src/features/playground/ui/PlaygroundDeck/PlaygroundDeck.tsx
memebattle-dev: Verified: two pile clicks before the server echo dispatch two identical open-stack placement commands. 3b5d4f427d513118fb1b418c7eb0c207f9083a2f adds the isValid guard for invalid piles. Earlier review feedback https://github.com/MemeBattle/monorepo/pull/686#discussion_r3912792974 assigned click clearing to the document. Should successful placement now clear selection immediately as an explicit exception, or retain focus with an in-flight guard? The double-click case remains unfixed pending that choice.

apps/ligretto-frontend/src/features/cardInteraction/ui/useDraggableCard.ts
memebattle-dev: Fixed in 3b5d4f427d513118fb1b418c7eb0c207f9083a2f: useCardInteraction is now the sole source identity/unmount selection-cleanup owner; useDraggableCard only owns native registration. Provider disablement already clears centrally. Regression tests observed two clear calls before the change and exactly one afterward, with native stale-source and disablement tests still green. All 110 frontend tests and 239 repository tests pass; frontend typecheck/build, formatting and lint pass.

apps/ligretto-frontend/src/entities/card/ui/Card/Card.tsx
memebattle-dev: 3b5d4f427d513118fb1b418c7eb0c207f9083a2f documents the current release-time activation, non-draggable latency and threshold-crossing cancelled placement in the README; the PR description is synchronized. LIG-144 deliberately used mousedown. Should we retain the current onClick trade-off, or restore mousedown activation for non-draggable cards? I left behavior unchanged rather than choose silently.

memebattle-dev:
Synced this branch with current master in bdf860b, resolving the lockfile conflict without changing Ligretto behavior. Preserved the DnD dependencies and updated the shared-package Vitest lockfile reference to match the base toolchain.

Validation: frozen-lockfile installation, repository unit tests, formatting/lint, Ligretto typechecks and frontend build passed; independent review passed. Repository-wide typechecking still reproduces the same three generated Blog route-type errors as an independently installed master after its build. CI has been restarted. The previously raised interaction-behavior questions remain unchanged and await maintainer direction.

apps/ligretto-frontend/src/features/cardInteraction/ui/useDroppableTarget.ts
```diff
@@ -0,0 +1,37 @@
...
+  const data = useMemo<CardDropData>(
+    () => ({
+      target,
+      onDrop: dragged => {
+        const state = store.getState()
+        const card = cardByInteractionTarget(state, dragged.target)
+        const destination = playgroundDecksSelector(state)[target.index]
+        if (enabled && card?.color === dragged.card.color && card?.value === dragged.card.value && canPlaceCardOnDeck(card, destination)) {
+          onDrop(dragged)
+        }
```
StepanDyubin, line 30: That should be defined in a component that uses the hook.

memebattle-dev: Addressed in 05b2201: PlaygroundDeck now validates the live Redux source identity and destination before dispatching a drop. useDroppableTarget keeps isValid highlighting and the provider enablement guard, but leaves placement acceptance to its consumer. Added boundary and live-state regressions. All 116 frontend tests and 250 repository tests pass; frontend typecheck/build, formatting, lint and independent review pass. Repository-wide typecheck still reproduces the same three generated Blog route-type errors on the unchanged baseline.

apps/ligretto-frontend/src/features/cardInteraction/ui/CardInteractionProvider.tsx
```diff
@@ -0,0 +1,130 @@
...
+    return () => document.removeEventListener('click', listener)
+  }, [state.mode])
+
+  const value = useMemo(
+    () => ({ activeTarget, clearActiveTarget, toggleActiveTarget, enabled, cardByTargetSelector }),
+    [activeTarget, clearActiveTarget, toggleActiveTarget, enabled, cardByTargetSelector],
+  )
+  return (
+    <CardInteractionContext value={value}>
+      <div style={{ display: 'contents' }} inert={!enabled}>
```
StepanDyubin: remove it

Issue #665, opened 2026-08-30
cas: persist passkeys, registration backed by DB


Themezv:
Table `passkey_credentials` (`account_id` FK, credential data, `name`, `created_at`, `last_used_at`). Registration ceremony creates an account (display name from request, replaces hardcoded "testuser") and stores the passkey on finish. Ceremony state stays in-memory with a TTL — single-instance limitation, note it in code.

Depends on #664.

**Acceptance criteria**
- Finished registration → account + credential rows.
- Stale ceremony state expires (TTL) instead of leaking.
- Registration start/finish covered by an integration test.

Issue #664, opened 2026-08-30
cas: accounts table and domain model


Themezv:
Migration + domain model for accounts: `id` (uuid, becomes OIDC `sub`), `display_name` (non-unique), `type` (`guest` | `full`), `email` (nullable, unverified for now), `created_at`, `last_seen_at`. Repository layer, no endpoints yet.

Per PLAN.md: no unique username, guests are regular rows.

**Acceptance criteria**
- Migration creates the table; Rust domain type + repository with create/get.
- Unit test covering the repository.

PR #688, opened 2026-09-01
feat(cas): accounts table and domain model (#664)

Themezv:
Depends on #689.

Adds the `accounts` table — the identity root, whose `id` is the OIDC `sub` — plus the Rust domain type and a repository with create/get. No endpoints yet; #665 wires registration on top.

Per docs/PLAN.md there is no unique username (`display_name` is non-unique) and guests are regular rows (`type = guest`), so guests and full accounts share one code path. `email` is nullable and unverified in v1; it is not unique yet because uniqueness without verification would let an unverified address block the real owner.

Repository queries use compile-time-checked SQLx macros and commit their offline cache. Tests use `#[sqlx::test]`, which gives each test a throwaway database with the migrations applied, so the CI test job runs against a Postgres service.

### DIFF
--- a/apps/cas/AGENTS.md
+++ b/apps/cas/AGENTS.md
@@ -8,15 +8,5 @@ Database migration workflow (sqlx, `cas-migrate`, immutability, expand/contract)
 SQL query and offline cache workflow is described in
 [docs/QUERIES.md](./docs/QUERIES.md).
 
-## Tests
-
-```
-cd apps/cas
-docker compose up -d                                    # once, for the dev DB
-DATABASE_URL=postgres://cas:cas@localhost:5434/cas cargo test
-```
-
-Repository tests use `#[sqlx::test]`, which needs `DATABASE_URL` in the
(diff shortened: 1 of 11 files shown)


bondiano (approved):

apps/cas/AGENTS.md
```diff
@@ -7,3 +7,16 @@ Database migration workflow (sqlx, `cas-migrate`, immutability, expand/contract)
 
 SQL query and offline cache workflow is described in
 [docs/QUERIES.md](./docs/QUERIES.md).
+
+## Tests
```
Themezv: @memebattle-dev, split this section to separated doc. Here only short description with link

memebattle-dev: Moved the detailed test setup and SQLx isolation notes to `apps/cas/docs/TESTS.md`; `AGENTS.md` now contains only a short description and link. Pushed in 7ef23e1. `git diff --check` passes.

Issue #666, opened 2026-08-30
cas: passkey login ceremony


Themezv:
`POST /api/webauthn/login-options` + `POST /api/webauthn/verify-login` using discoverable credentials — no identifier asked. Update signature counter and `last_used_at` on success.

Depends on #665.

**Acceptance criteria**
- An account registered via #665 can log in.
- Unknown/invalid credential → 401, no panic.
- Integration test for the ceremony.

PR #691, opened 2026-09-03
feat(cas): persist passkeys, registration backed by DB (#665)

### LINKED ISSUE #665: cas: persist passkeys, registration backed by DB

Themezv:
Registration is now backed by the database: the display name comes from the
request (replacing the hardcoded `testuser`), and a finished ceremony creates
the account together with its first passkey. Along the way the crate got the
layout it will keep growing in: `docs/LAYOUT.md`.

## Decisions

- **`passkey_credentials`** stores the webauthn-rs `Passkey` whole as `jsonb`,
  with the raw `credential_id bytea UNIQUE` lifted into its own column — login
  (#666) starts from that id and needs an index on it. No `sign_count` column;
  the counter lives in the credential. ADR:
  `apps/cas/docs/adr/0001-passkey-persistence.md`.
- **Ceremony state lives in Postgres** (`webauthn_ceremonies`), never in
  process memory: CAS runs as autoscaled replicas, so the two requests of one
  ceremony may land on different pods. A row is single-use — consuming it
  deletes it in the same statement — and lives for the browser timeout plus a
  grace period measured by the database clock. webauthn-rs's
  `danger-allow-state-serialisation` feature is what allows this; the danger it
  guards against is client-side storage, which CAS never does. Expired rows are
  not swept on the request path; that is a separate job. ADR:
  `apps/cas/docs/adr/0002-ceremony-state-in-postgres.md`.
- **The account id is minted by CAS**, not by the database: it goes to the
  authenticator as the WebAuthn user handle before the row exists, is baked
  into the credential and can never change. `NewAccount` carries it,
  `with_id` pins a chosen one.
- **Finishing is one transaction**: consume the ceremony, verify, write the
  account and the credential. A failed verification commits the consumed
  ceremony (the challenge was answered, wrongly); a failed write rolls
  everything back so the client can retry. An abandoned registration leaves
  nothing but an expiring ceremony row.
- **`DisplayName` is a validated newtype** (nutype): Unicode spaces mapped to
  ASCII, NFC, trimmed and collapsed; invisible, control and bidi characters
  rejected using Unicode categories and ICU default-ignorable properties;
  marks, selectors and joiners require a base, preserving emoji ZWJ sequences;
  ≤ 64 characters. Decoding a row
  re-validates, so a name written outside CAS that does not pass fails to
  decode instead of reaching a UI. `400 invalid_display_name` on the wire.
  `precis-profiles` was tried and dropped (panics on non-ASCII trimming,
  rejects emoji).
- **Database failures are classified once**, in `db`: an outage is `503
  database_unavailable`, a serialization failure or deadlock is `503
  database_busy`, and anything the code has no name for stays a 500. A
  transport only maps the verdict to a status.
- **Discoverable credentials are required** for the original usernameless
  sign-in flow: registration sends `residentKey: required`,
  `requireResidentKey: true`, and `userVerification: required`. CAS adapts the
  public creation options while retaining webauthn-rs's high-level verifier
  and support for synced passkeys. An explicit `credProps.rk: false` returns
  `400 discoverable_credential_required`; the optional, unsigned extension
  is not treated as authentication proof. Old-policy pending ceremonies are
  discarded and must restart. ADR 0001 and PLAN now document this requirement,
  replacing the earlier non-discoverable-credential fallback.
- **The finish response returns `accountId` and `credentialId`**, not the
  credential.

## Layout

One directory per bounded context (`accounts`, `webauthn`), with the domain
in its own files, services orchestrating transactions, every query of the
context in `repository.rs` — the only place with SQL and hand-written sqlx
impls — and the context's handlers in `<context>/http/`. The crate-level
`http/` is the transport root: middleware, `ApiError`, extractors, and the
mounting of each context's router. `config`, `db`, `migrations` are shared
infrastructure that knows no context.
The rules are in `apps/cas/docs/LAYOUT.md`; `AGENTS.md` is now a card of
links.

## Tests

Repository, service and handler tests use a resident-storage test adapter over
`webauthn-authenticator-rs`'s `SoftPasskey`, retaining real key generation and
signing. The adapter emulates RP-scoped discovery and user handles; it is not
a browser or hardware compatibility test. Covered: the full start → finish round trip
creating both rows, a replayed finish (404), display name sanitising and
validation, the jsonb round trip, the unique-credential rollback, the cascade
delete, ceremony single-use / expiry / kind mismatch / rollback, an undecodable
ceremony being discarded, and the invalid-row decode guard. Unicode regression
tests cover invisible-only names (CGJ, selectors, Hangul fillers and blank
Braille), invalid selector/joiner placement, and preservation of emoji,
combining marks, CJK and Mongolian variation selectors. The rejection is also
checked through serde, HTTP validation and database decoding. Registration
tests cover required resident-key options, unsupported authenticators,
negative and missing credential properties, old-policy ceremony invalidation,
and usernameless authentication against a credential read back from Postgres.

```
cd apps/cas
docker compose up -d
DATABASE_URL=postgres://cas:cas@localhost:5434/cas cargo test
SQLX_OFFLINE=true cargo clippy --all-targets -- -D warnings
cargo fmt --check
DATABASE_URL=postgres://cas:cas@localhost:5434/cas cargo sqlx prepare --check
```

78 tests pass. Formatting and Clippy with warnings denied also pass. The manual browser client in `apps/cas/tests/client` has a
display-name input, explains unsupported passkeys, and reports the created account id.
Its TypeScript check and production build pass.

Closes #665


apps/cas/AGENTS.md
```diff
@@ -10,3 +24,6 @@ SQL query and offline cache workflow is described in
 
 Test setup and the isolated SQLx database workflow are described in
 [docs/TESTS.md](./docs/TESTS.md).
+
+Decisions that shaped the schema and the ceremonies are recorded as ADRs in
```
Themezv, line 28: Remove it

apps/cas/AGENTS.md
```diff
@@ -2,6 +2,20 @@
 
 Rust (axum) authentication service.
 
+## Layout
```
Themezv, line 5: Move to docs. Here link only. Review structure

PR #701, opened 2026-09-12
feat(cas): Fetch Metadata / Origin check on mutating requests (#696)

### LINKED ISSUE #696: cas: Fetch Metadata / Origin check on mutating requests (CSRF)
OWASP CSRF cheat sheet: `SameSite` is defense in depth, not a standalone defence. Our JSON endpoints are covered by the CORS preflight, but `POST /api/logout` takes no body and any future body-less mutating endpoint (passkey delete, #668) would rely on `SameSite=Lax` alone.

Add a tower layer on `/api` that rejects a mutating request (anything but `GET`/`HEAD`/`OPTIONS`) unless:

- `Sec-Fetch-Site` is `same-origin`, `same-site` or `none`; or
- `Sec-Fetch-Site` is `cross-site` and `Origin` is in `CAS_CORS_ORIGINS`; or
- `Sec-Fetch-Site` is absent (old browser / non-browser client) and `Origin` is absent or in `CAS_CORS_ORIGINS`.

`GET` navigations stay open: `/authorize` (M2) arrives as `cross-site` + `navigate`.

Rejection: `403 cross_site_request`, standard `ApiError` shape. Decision and the fallback rules go into an ADR.

Depends on #667.

**Acceptance criteria**
- Cross-site form `POST /api/logout` with a valid cookie → 403, session untouched.
- `fetch` from an allowed origin (`Sec-Fetch-Site: cross-site` or `same-site`, `Origin` allowed) → passes.
- Request without any Fetch Metadata and without `Origin` → passes.
- `GET` from anywhere → untouched by the layer.
- Router tests cover each branch.

Themezv:
Closes #696.

## What

- `src/http/fetch_metadata.rs`: a layer on the `/api` router that refuses a mutating request (anything but `GET`/`HEAD`/`OPTIONS`) unless
  - `Sec-Fetch-Site` is `same-origin` or `none`; or
  - `Sec-Fetch-Site` is `same-site`, `cross-site` (or any unknown value) and `Origin` is in `CAS_CORS_ORIGINS`; or
  - `Sec-Fetch-Site` is absent and `Origin` is absent or in `CAS_CORS_ORIGINS`.
- `same-site` is held to the origin list because a sibling subdomain shares the cookie jar and can post an HTML form that no CORS preflight sees (review finding).
- Refusal: `403 cross_site_request`, standard `ApiError` shape, before the handler and before any DB access. Reason logged at `warn` with method, path and both headers.
- `ApiError::forbidden` constructor.
- `/api` is now built as one router (`api_router`) and wrapped by the layer before being nested, so every route and the fallback under it are covered. `/health` stays outside.
- ADR 0005 records the decision and the fallback rules; ADR 0004 (e) and its consequences point at it. `docs/LAYOUT.md` lists the new file.

## Tests

- Pure `verdict` tests for each branch: same-origin/none, same-site and cross-site with/without allowed origin, unknown `Sec-Fetch-Site` value, no Fetch Metadata with absent/allowed/other/`null` origin, safe vs. mutating methods, exact origin comparison.
- Router tests: cross-site form `POST /logout` with a valid cookie → 403, no `Set-Cookie`, session still answers `/me`; same-site form from a sibling subdomain → 403; `fetch` from the allowed origin logs out; request without any headers passes; `GET` from anywhere untouched; both nested contexts are behind the layer; the fallback is guarded.

`cargo fmt --check`, `cargo clippy --all-targets --all-features` and `cargo test` (147 tests, against the dev Postgres) are clean.

### DIFF
--- a/apps/cas/src/http/fetch_metadata.rs
+++ b/apps/cas/src/http/fetch_metadata.rs
@@ -1,0 +1,346 @@
+//! The CSRF line for `/api`: a mutating request must come from this site or
+//! from an origin the frontend is served from. Decided in ADR 0005.
+//!
+//! The check reads the browser's Fetch Metadata (`Sec-Fetch-Site`) and falls
+//! back to `Origin` for clients that send neither. It costs no state and no
+//! token: the browser states where a request came from and the server says
+//! whether that is good enough.
+
+use std::sync::Arc;
+
+use axum::{
+    Router,
+    extract::{Request, State},
+    http::{HeaderMap, HeaderValue, Method, header},
+    middleware::{self, Next},
+    response::Response,
+};
+
+use crate::http::error::ApiError;
+
+const SEC_FETCH_SITE: &str = "sec-fetch-site";
+
+/// The origins a cross-site mutating request may come from: the frontend in
+/// development and any other site CORS already admits. Shared by every
+/// request, hence the `Arc`.
+#[derive(Debug, Clone)]
+pub struct AllowedOrigins(Arc<[HeaderValue]>);
+
+impl AllowedOrigins {
+    pub fn new(origins: Vec<HeaderValue>) -> Self {
+        Self(origins.into())
+    }
+
+    fn contains(&self, origin: &HeaderValue) -> bool {
+        self.0.iter().any(|allowed| allowed == origin)
+    }
+}
+
+/// Wraps every route of `router`, including its fallback, in the check. The
+/// `/api` router goes through here before it is mounted.
+pub fn guard(router: Router, origins: AllowedOrigins) -> Router {
+    router.layer(middleware::from_fn_with_state(origins, reject_cross_site))
+}
+
+async fn reject_cross_site(
+    State(origins): State<AllowedOrigins>,
+    request: Request,
+    next: Next,
+) -> Result<Response, ApiError> {
+    verdict(request.method(), request.headers(), &origins).map_err(|rejection| {
+        tracing::warn!(
+            method = %request.method(),
+            path = request.uri().path(),
+            sec_fetch_site = ?request.headers().get(SEC_FETCH_SITE),
+            origin = ?request.headers().get(header::ORIGIN),
+            reason = rejection.reason(),
+            "cross-site mutating request rejected"
+        );
+        ApiError::forbidden("cross_site_request", "Cross-site requests are not allowed")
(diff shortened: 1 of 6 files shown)


apps/cas/src/http/fetch_metadata.rs
```diff
@@ -0,0 +1,346 @@
...
+
+    let origin_allowed = || {
+        headers
+            .get(header::ORIGIN)
+            .is_some_and(|origin| origins.contains(origin))
+    };
+
+    match headers.get(SEC_FETCH_SITE).map(HeaderValue::as_bytes) {
+        Some(b"same-origin" | b"same-site" | b"none") => Ok(()),
+        Some(_) if origin_allowed() => Ok(()),
```
Themezv, line 112: [P2] Require an allowed Origin for same-site mutations

A page on an untrusted or compromised sibling subdomain can submit a form to `/api/logout` with `Sec-Fetch-Site: same-site`. The browser includes the session cookie, and this branch accepts the request even when its Origin is absent from `CAS_CORS_ORIGINS`, allowing forced logout. CORS does not prevent ordinary HTML form submissions, and the planned `__Host-` prefix prevents cookie overwrites rather than this attack.

Please require an allowed Origin for same-site mutations as well; the legitimate sibling frontend can use the existing allowlist. This follows [OWASP's guidance on sibling-domain trust](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#how-to-treat-fetch-metadata-headers-on-the-server-side). The current unconditional allowance is appropriate only if all sibling subdomains are explicitly within the trusted boundary.

Themezv: Agreed, and fixed in 369d959f. `same-site` now sits in the same branch as `cross-site`: a mutating request passes only with an `Origin` from `CAS_CORS_ORIGINS`. `same-origin` and `none` still pass unconditionally, since neither can come from a third party.

The frontend is unaffected wherever it is served from: `localhost:5173` → `localhost:3000` is reported as `same-site` and its origin is already in the list. Added a test for a form post from a sibling subdomain (`same-site`, foreign `Origin` → 403) and rewrote ADR 0005 (b) with this reasoning; the old argument about `__Host-` covering siblings was wrong, as you note.

Issue #698, opened 2026-09-12
cas: Cache-Control: no-store on authenticated responses, Clear-Site-Data on logout


Themezv:
OWASP Session Management: responses carrying session-bound data must not be cached; logout should ask the browser to drop cached data and cookies.

- `Cache-Control: no-store` on every `/api` response (simplest: a layer on the `/api` router; nothing under it is cacheable).
- `Clear-Site-Data: "cache", "cookies", "storage"` on `POST /api/logout`. Note: `"cookies"` clears every cookie for the site, which is what we want on the CAS origin; `"storage"` is fine because CAS stores nothing client-side.

Depends on #667.

**Acceptance criteria**
- `GET /api/me` and the webauthn endpoints answer with `Cache-Control: no-store`.
- `POST /api/logout` answers with `Clear-Site-Data` and still clears the session cookie explicitly.
- Router tests for both headers.

PR #702, opened 2026-09-12
feat(cas): idle timeout with sliding renewal for sessions (#697)

### LINKED ISSUE #697: cas: decide session timeouts (idle vs absolute)
ADR 0004 gives sessions a single absolute lifetime of 30 days and a persistent cookie (`Max-Age`). OWASP Session Management recommends an idle timeout on top of an absolute one and a non-persistent cookie for session tracking; typical values there are far shorter. Our product (SSO for a game) may justify long sessions, but the choice must be explicit.

Decide and record in ADR 0004 one of:

- keep 30 days absolute, persistent cookie, with the reasoning; or
- add an idle timeout with sliding renewal (a write at most once per N minutes, never on every request) and keep an absolute cap.

If renewal is chosen: `last_seen_at` on `sessions`, the renewal window as a constant next to `SESSION_LIFETIME`, cookie `Max-Age` re-sent on renewal.

Depends on #667.

**Acceptance criteria**
- ADR 0004 states the chosen timeouts and why.
- If idle timeout: repository/service tests for renewal and for expiry after inactivity; no write on requests inside the renewal window.

Themezv:
Closes #697.

## Decision (ADR 0004, amended in (c))

Sessions get two clocks, both the database's:

- **Idle timeout, 7 days**, reset by an authenticated request at most once per **renewal window of 1 hour**. A request inside the window writes nothing; one outside it resets `last_seen_at` on the session and on the account in one transaction.
- **Absolute cap, 30 days**, unchanged: `expires_at` is fixed at creation and never moves.

Why seven days: it is what a player who opens the game most weeks never notices, and a passkey makes signing in again a single touch, which is what makes a shorter idle limit affordable. The cap bounds how long a stolen cookie is worth anything. The cookie stays persistent (`Max-Age`): players close the tab between rounds.

## Changes

- Migration: `sessions.last_seen_at` (default `now()`, nobody is signed out) and an index for the future cleanup job.
- `find_live` checks both clocks and reports whether renewal is due, all by the database clock; `renew` resets the idle clock only.
- `SessionService::authenticate` returns `(Authenticated, Renewal)`; renewal is one transaction with the account's `last_seen_at`.
- Cookie `Max-Age` runs to `Session::valid_until()` (idle timeout from the last reset, or the cap, whichever is first). A renewal re-sends the cookie: the `Authenticated` extractor cannot touch the response, so a layer on `/api` (`sessions::http::renewal`) carries it out through a per-request slot. A handler that sets the cookie itself wins.
- `GET /api/me` reports `sessionExpiresAt` as that effective moment rather than the absolute cap.
- `.sqlx` refreshed.

## Tests

Repository: idle expiry is not found and not removed, renewal due only outside the window, renew keeps the cap. Service: no write inside the window, renewal outside it moves session and account, idle session does not authenticate. Transport: renewal re-sends the cookie with `Max-Age` at the idle timeout, no cookie inside the window, handler's own cookie wins, no cookie on 401.

`cargo test`: 160 passed. `cargo clippy --all-targets -- -D warnings`: clean.

## Note

#698 (in flight) also touches `api_router` and ADR 0004; whichever merges second needs a small rebase.

### DIFF
--- a/apps/cas/.sqlx/query-23a054955b2f55cce14164c2081966497e1e808394e91af99e5637221154c287.json
+++ b/apps/cas/.sqlx/query-23a054955b2f55cce14164c2081966497e1e808394e91af99e5637221154c287.json
@@ -1,6 +1,6 @@
 {
   "db_name": "PostgreSQL",
-  "query": "UPDATE sessions SET last_seen_at = now() WHERE id = $1 RETURNING last_seen_at",
+  "query": "UPDATE sessions SET last_seen_at = now()\n           WHERE id = $1 AND last_seen_at <= now() - make_interval(secs => $2)\n           RETURNING last_seen_at",
   "describe": {
     "columns": [
       {
@@ -17,12 +17,13 @@
     ],
     "parameters": {
       "Left": [
-        "Uuid"
+        "Uuid",
+        "Float8"
       ]
     },
     "nullable": [
       false
     ]
   },
-  "hash": "c935d10d0dc39d07738550cb5e418a8a71ce2b0a5b9bb4bd2eaf24bff3bd5831"
+  "hash": "23a054955b2f55cce14164c2081966497e1e808394e91af99e5637221154c287"
 }
--- a/apps/cas/.sqlx/query-c935d10d0dc39d07738550cb5e418a8a71ce2b0a5b9bb4bd2eaf24bff3bd5831.json
+++ b/apps/cas/.sqlx/query-c935d10d0dc39d07738550cb5e418a8a71ce2b0a5b9bb4bd2eaf24bff3bd5831.json
@@ -1,0 +1,28 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "UPDATE sessions SET last_seen_at = now() WHERE id = $1 RETURNING last_seen_at",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "last_seen_at",
+        "type_info": "Timestamptz",
+        "origin": {
+          "Table": {
+            "table": "sessions",
+            "name": "last_seen_at"
+          }
+        }
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Uuid"
+      ]
+    },
+    "nullable": [
+      false
+    ]
+  },
+  "hash": "c935d10d0dc39d07738550cb5e418a8a71ce2b0a5b9bb4bd2eaf24bff3bd5831"
+}
--- a/apps/cas/.sqlx/query-d959a86751b33822ea19aa36e09d612f348c493eebe87460738644c58ab5e8b3.json
+++ b/apps/cas/.sqlx/query-d959a86751b33822ea19aa36e09d612f348c493eebe87460738644c58ab5e8b3.json
@@ -1,6 +1,6 @@
 {
   "db_name": "PostgreSQL",
-  "query": "SELECT id, account_id, created_at, expires_at\n           FROM sessions\n           WHERE token_hash = $1 AND expires_at > now()",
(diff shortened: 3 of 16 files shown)


Themezv:
Review findings for 23b77098277134452bd3bb79dcc58c4ee2c1d545:

1. **[P2] Handle a session deleted during renewal as unauthenticated.**

   In [`sessions/repository.rs:91–95`](https://github.com/MemeBattle/monorepo/blob/23b77098277134452bd3bb79dcc58c4ee2c1d545/apps/cas/src/sessions/repository.rs#L91-L95), `renew` uses `fetch_one`. If logout deletes the session after `find_live` reads it but before renewal completes, the update returns no row and produces `RowNotFound`. This propagates through `authenticate` into HTTP 500 instead of the expected 401. Return an optional renewal result and map a missing session to `Ok(None)`.

   Reproduced deterministically by holding an uncommitted session deletion, starting authentication, waiting for its renewal update to block, and committing the deletion: authentication returned `Err(RowNotFound)`.

2. **[P2] Enforce the renewal window atomically.**

   In [`sessions/service.rs:80–85`](https://github.com/MemeBattle/monorepo/blob/23b77098277134452bd3bb79dcc58c4ee2c1d545/apps/cas/src/sessions/service.rs#L80-L85), the renewal decision comes from an earlier read outside the transaction. Concurrent requests can all read `renewal_due = true`, then each unconditionally updates the session and account. This defeats the once-per-hour write limit and serializes redundant writes on the same rows. Recheck the renewal window atomically in the database, and only touch the account when renewal actually occurs; requests that lose the renewal race should still authenticate using the current session state.

   Reproduced by locking a renewal-due session, starting two authentication requests, waiting for both renewal updates to block, and releasing the lock: both returned `Renewal::Renewed`.

Also, the documented idle-time range is reversed: with hourly timestamp updates, expiry occurs between **7 days minus 1 hour and 7 days after the last authenticated request**, not between 7 days and 7 days plus 1 hour (and the absolute cap can shorten this further). Please correct ADR 0004 and the `SESSION_RENEWAL_WINDOW` comment.

Validation: all 160 existing CAS tests passed against the local PostgreSQL service with `SQLX_OFFLINE=true`. Both temporary concurrency regression checks reproduced the findings above; the temporary test file was removed after the review.

Themezv:
All three findings are right and are fixed in 
the follow-up commit on this branch.

**1 and 2, one change.** `renew` now carries the window check in the `UPDATE` itself (`WHERE id = $1 AND last_seen_at <= now() - window`) and returns `Option`. Two requests that both read the clock as due take the row lock in turn; the second re-evaluates the condition on the row the first just wrote and matches nothing. `authenticate` treats a `None` from the write as "arrived a moment later": it reads the session again and authenticates with what it finds (`Renewal::Kept`), or answers `Ok(None)` if logout removed it in between. The account's `last_seen_at` is touched only in the branch where the session's actually moved, in the same transaction as before.

Both scenarios are now regression tests in `service.rs`, built the way you reproduced them: `concurrent_requests_renew_once` holds the row with `SELECT ... FOR UPDATE`, spawns two `authenticate` calls, releases, and expects exactly one `Renewed` with both callers seeing the same `last_seen_at`; `a_session_deleted_during_renewal_does_not_authenticate` holds an uncommitted `DELETE`, spawns one, commits, and expects `Ok(None)`. Plus a repository test that `renew` writes nothing inside the window or for a missing id.

**3.** Corrected in the `SESSION_RENEWAL_WINDOW` comment and in ADR 0004 (c): the clock lags the last request by up to an hour, so measured from that request a session ends between seven days less an hour and seven days (the cap can cut it shorter still).

`cargo test`: 163 passed. The two race tests were run five times in a row without a flake.

PR #703, opened 2026-09-12
feat(cas): Cache-Control: no-store on /api, Clear-Site-Data on logout (#698)

### LINKED ISSUE #698: cas: Cache-Control: no-store on authenticated responses, Clear-Site-Data on logout

Themezv:
Closes #698.

## Changes

- **`Cache-Control: no-store` on every `/api` response.** A `SetResponseHeaderLayer` on the `/api` router, outside the CSRF line, so that accounts, ceremony steps, validation errors and the guard's own 403s all carry it. `/health` and paths outside `/api` are untouched.
- **`Clear-Site-Data: "cache", "cookies", "storage"` on `POST /api/logout`**, next to the explicit removal cookie that stays for browsers without `Clear-Site-Data` support. `cookies` clears every cookie on the CAS origin, which is what logout wants; `storage` costs nothing because CAS stores nothing client-side.
- **The `/api` router owns a fallback.** Found while wiring the layer: a nested router without a fallback hands unknown paths to the outer router, where neither the CSRF line nor `no-store` can see them, so `POST /api/nowhere` cross-site answered a bare 404 instead of the 403 ADR 0005 promises. With its own fallback, an unknown `/api` path is `404 not_found` in the `ApiError` shape and sits behind every layer. ADR 0005 (a) and the `guard` doc comment now say why, and the fallback test goes through the real `app()` composition.
- ADR 0004 gains decision (i).

Rebased onto #702 (idle timeout): the cookie-renewal layer, the CSRF guard and `no-store` now stack on the same router.

## Tests

`no-store` on `GET /api/me` (200 and 401), a webauthn 422, the guard's 403 and the fallback 404; nothing outside `/api` marked. `Clear-Site-Data` on logout with and without a session, removal cookie still present. Unknown `/api` and `/api/webauthn` paths: 404 `not_found` JSON, and 403 `cross_site_request` when probed cross-site.

`cargo test`: 169 passed. `cargo clippy --all-targets -- -D warnings`: clean.

### DIFF
--- a/apps/cas/src/http/mod.rs
+++ b/apps/cas/src/http/mod.rs
@@ -14,10 +14,12 @@ use axum::{
 };
 use sqlx::postgres::PgPoolOptions;
 use thiserror::Error;
+use tower::Layer;
 use tower::ServiceBuilder;
 use tower_http::{
     catch_panic::CatchPanicLayer,
     cors::{AllowOrigin, CorsLayer},
+    normalize_path::{NormalizePath, NormalizePathLayer},
     set_header::SetResponseHeaderLayer,
     trace::{self, TraceLayer},
 };
@@ -56,7 +58,16 @@ pub struct ApiState {
     pub cookies: CookieSettings,
 }
 
-pub fn app(config: Config) -> Result<Router, AppError> {
+/// The whole service: the routers behind their middleware, with trailing
+/// slashes trimmed before routing. `NormalizePath` wraps the `Router` from
+/// the outside rather than through `Router::layer`, because routing has
+/// already happened by the time a `Router::layer` middleware runs; from the
+/// outside, `/api/` becomes `/api` and lands on the `/api` router's own
+/// fallback, behind its layers, instead of falling through to the outer
+/// router (a nested router's catch-all does not match an empty rest, so the
+/// slash-terminated prefix alone escaped it). `/api/me/` is `/api/me`, and
+/// nothing served here gives a trailing slash a meaning of its own.
+pub fn app(config: Config) -> Result<NormalizePath<Router>, AppError> {
     // Lazy pool: connections open on first use, so startup succeeds even when
     // the DB is down and `/health` reports the actual connectivity.
     let pool = PgPoolOptions::new()
@@ -78,7 +89,8 @@ pub fn app(config: Config) -> Result<Router, AppError> {
         api_router(api_state, AllowedOrigins::new(config.cors_origins.clone())),
     );
 
-    Ok(with_middleware(router, config.cors_origins))
+    Ok(NormalizePathLayer::trim_trailing_slash()
+        .layer(with_middleware(router, config.cors_origins)))
 }
 
 /// Everything under `/api`: the contexts' routers, re-sending the session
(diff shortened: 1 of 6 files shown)


Themezv (commented):
Reviewed e541444a. All 169 CAS tests passed against an isolated PostgreSQL database. Additional router probes reproduced the `/api/` fallback gap described below.

apps/cas/src/sessions/http/mod.rs
```diff
@@ -35,6 +41,21 @@ impl From<CreateError> for ApiError {
...
+/// of the value: the header carries a list of quoted directives.
+///
+/// `cache` is the point of the exercise: no cached answer about the account
+/// that just signed out may be replayed by the back button. `cookies` clears
+/// every cookie of the site, which is the right blast radius here — the CAS
+/// origin serves nothing but CAS, so the only cookie to lose is the session.
+/// `storage` costs nothing, because CAS stores nothing client-side, and
+/// covers whatever a frontend on this origin may leave behind later.
+const CLEAR_SITE_DATA_ON_LOGOUT: HeaderValue =
+    HeaderValue::from_static(r#""cache", "cookies", "storage""#);
```
Themezv: [P2] Account for domain-wide cookie deletion on logout

`Clear-Site-Data: "cookies"` clears cookies across the entire registered domain, including host-only cookies on sibling subdomains; it is not scoped to the CAS origin. For example, a response from `cas.mems.fun` also removes any cookies belonging to `mems.fun`, `blog.mems.fun`, and other siblings. This happens even for the idempotent logout without a CAS session, so it can remove other applications' sessions or preferences when those cookies exist. See the [specification's cookie-clearing algorithm](https://w3c.github.io/webappsec-clear-site-data/#clear-cookies).

This implements the header explicitly requested by #698, but the requirement and the comment above rely on the incorrect assumption that a dedicated CAS origin limits cookie deletion to CAS. If domain-wide clearing is intended, document that scope and consequence in the comment and ADR. If logout should only clear the CAS session cookie, omit the `cookies` directive and retain the explicit removal cookie.

apps/cas/src/http/mod.rs
```diff
@@ -82,14 +83,41 @@ pub fn app(config: Config) -> Result<Router, AppError> {
...
+/// a ceremony challenge, a validation error, a 403 from the CSRF line, a
+/// path that does not exist — and a new endpoint must be covered by where it
+/// is mounted. It wraps the CSRF line from the outside so that the layer's
+/// own refusals carry it too.
 fn api_router(state: ApiState, origins: AllowedOrigins) -> Router {
     let api = Router::new()
         .nest("/webauthn", webauthn_http::router(state.clone()))
-        .merge(sessions_http::router(state));
+        .merge(sessions_http::router(state))
+        .fallback(not_found);
```
Themezv, line 112: [P3] Include `/api/` in the guarded fallback

The nested fallback still misses the trailing-slash root `/api/`. Through the real `app()` composition, I reproduced `GET /api/` returning a bare 404 without `Cache-Control: no-store`, and `POST /api/` with `Sec-Fetch-Site: cross-site` and `Origin: https://evil.example` returning 404 instead of the intended 403. `/api` and `/api/nowhere` behave correctly.

Route `/api/` through the same API middleware and include both requests in the composition tests. This is a gap in the PR's stated response contract, not a bypass that reaches a protected handler or exposes session data.

apps/cas/src/sessions/http/mod.rs
Themezv: Right, and thank you for checking the algorithm rather than the ticket text. The spec clears cookies for the registrable domain of the response's host, so on `cas.mems.fun` the directive would have reached `mems.fun`, `blog.mems.fun` and every other sibling. That is not what a CAS logout means: ending sessions elsewhere is RP-initiated logout (M2), a deliberate act.

Fixed in 6cf6f1f7: the value is now `"cache", "storage"`, both origin-scoped; the explicit removal cookie removes the session cookie as before. The constant's doc comment says why `cookies` is absent, ADR 0004 (i) records the reasoning with the domain example, and the tests assert the new value. The ticket's assumption was wrong; #698 will be closed by this PR with the corrected scope.

apps/cas/src/http/mod.rs
Themezv: Reproduced. A nested router's catch-all (`/api/{*rest}`) does not match an empty rest, so `/api/` was the one path with the prefix that matched neither `/api` nor the nested fallback and fell through to the outer 404.

Rather than routing `/api/` by hand, 6cf6f1f7 trims trailing slashes before routing: `NormalizePathLayer::trim_trailing_slash()` wraps the `Router` from the outside (through `tower::Layer`, not `Router::layer`, since routing has happened by the time a `Router::layer` middleware runs), and `main.rs` builds the service factory through `axum::ServiceExt`. `/api/` is now `/api`, `/api/me/` is `/api/me`, and nothing served here gives a trailing slash a meaning of its own. The composition tests cover `GET /api/` (404 `not_found`, `no-store`), `POST /api/` cross-site (403 `cross_site_request`, `no-store`) and `GET /api/me/` (401, the route itself). ADR 0004 (i) mentions it next to the fallback.

Issue #700, opened 2026-09-12
cas: __Host- prefix for the session cookie on https origins


Themezv:
OWASP Session Management recommends the `__Host-` cookie name prefix: the browser then refuses the cookie unless it is `Secure`, has no `Domain` and `Path=/`, which is exactly how we set it (ADR 0004). It closes cookie tossing from sibling subdomains.

The prefix requires `Secure`, and `CookieSettings::secure` is false on the plain-http development origin, so the name must follow the setting:

- `__Host-cas_session` when `secure` is true;
- `cas_session` otherwise.

`SESSION_COOKIE` becomes a method on `CookieSettings`; the extractor and logout read the same name.

Depends on #667.

**Acceptance criteria**
- With an https `CAS_ORIGIN` the cookie is named `__Host-cas_session`, `Secure`, `Path=/`, no `Domain`.
- With the http dev origin the name stays `cas_session`.
- Extractor, logout and the removal cookie use the configured name; tests cover both settings.

PR #705, opened 2026-09-12
feat(cas): log the session lifecycle (#699)

### LINKED ISSUE #699: cas: log the session lifecycle
OWASP Session Management: log session creation, destruction and invalid-session activity, never the session id itself.

Add `tracing` events in `SessionService`:

- `info` on `create`: session id, account id, how it was created (registration or login).
- `info` on `revoke`: session id (looked up by hash before delete, or returned from `DELETE ... RETURNING id`), or `debug` when nothing was found.
- `warn` in the `Authenticated` extractor when a cookie with a well-formed token names no live session (unknown, expired or revoked); no event for a request with no cookie at all.

The token never appears in any event; the row id is the identity for logs (ADR 0004).

Depends on #667.

**Acceptance criteria**
- Events above emitted with the fields listed; a `tracing_test`-style or `tracing-subscriber` capture test proves the token is not in any event.

Themezv:
Closes #699.

## Events

| Where | Level | Event | Fields |
|---|---|---|---|
| `SessionService::create` | `info` | session created | `session_id`, `account_id`, `origin` (`registration` / `login`, a new `SessionOrigin` passed by the finish handlers) |
| `SessionService::authenticate` | `debug` | session renewed | `session_id`, `account_id` (the #697 renewal, at most hourly per session) |
| `SessionService::revoke` | `info` / `debug` | session revoked / logout without a live session | `session_id` from `DELETE ... RETURNING id` |
| `Authenticated` extractor | `warn` | session cookie names no live session | `path`; `debug` for a malformed cookie value; nothing for a request without a cookie |

The token, its hash and the cookie value appear in no event; the row id is the identity for logs (ADR 0004 (j)).

## The request trace was leaking the token

`TraceLayer` had `DefaultOnResponse::include_headers(true)`, which logged response headers at INFO, `Set-Cookie` with the raw session token included. Every registration, login and renewal wrote the token to the log. Confirmed empirically, then fixed by dropping `include_headers` (method, path, status and latency stay). A test runs a real registration through `with_middleware` and asserts the token is in no captured event.

## Capture test

`testing::capture_tracing()` installs a collecting layer on the current thread via `set_default`; `sqlx::test` runs a current-thread runtime, so the guard covers the async code. It also installs, once per process, an empty global `Registry`: `tracing` caches per callsite whether a line is enabled and decides it from the global subscriber, so without one a callsite first reached by another test's thread was disabled for the whole binary and the capture missed it silently. Documented in the helper and in `docs/TESTS.md`.

## Tests

Service and extractor events with their fields, the negative check that no captured text contains the token or its hash, the transport-level check above. `cargo test`: 168 passed (four full runs, no flakes). `cargo clippy --all-targets -- -D warnings`: clean. `.sqlx` refreshed for the `RETURNING id` delete.

Conflicts with #706 (#700) are small: the last Consequences bullet of ADR 0004 and `testing::session_cookie`.


apps/cas/src/sessions/http/extract.rs
```diff
@@ -39,16 +45,32 @@ where
...
+        };
 
-        let (authenticated, renewal) = state
-            .sessions
-            .authenticate(&token)
-            .await?
-            .ok_or_else(unauthenticated)?;
+        let Some((authenticated, renewal)) = state.sessions.authenticate(&token).await? else {
+            tracing::warn!(
+                path = parts.uri.path(),
```
Themezv: [P2] Preserve the original request path

Axum strips nested router prefixes before this extractor runs, so an unknown or revoked token sent to GET /api/passkeys produces `session cookie names no live session path="/"`. I reproduced this through the nested API router. This loses the endpoint identity from the invalid-session warning; the default INFO logging configuration also filters out the DEBUG request span that could otherwise supply the full URI. Read the path from `OriginalUri` for this warning and the malformed-cookie event, and cover a nested route in the test.

apps/cas/src/testing.rs
```diff
@@ -50,6 +57,93 @@ pub fn session_cookie(response: &axum::response::Response) -> Option<SessionToke
...
+    let events = CapturedEvents::default();
+    let guard =
+        tracing::subscriber::set_default(Registry::default().with(CaptureLayer(events.clone())));
+    (events, guard)
+}
+
+struct CaptureLayer(CapturedEvents);
+
+impl<S: tracing::Subscriber> Layer<S> for CaptureLayer {
+    fn on_event(&self, event: &tracing::Event<'_>, _ctx: Context<'_, S>) {
```
Themezv, line 134: [P2] Include span fields in the secret-leak capture

This layer only records event fields and ignores enclosing span fields, while the production `tracing_subscriber::fmt` layer prints those fields with each event. A probe that entered a request span containing a secret in its `headers` field and emitted an INFO event captured only the message, so the negative secret check would pass even though production output would contain the secret. This leaves the new regression checks blind to request-span leaks, such as enabling request headers on `DefaultMakeSpan`. Capture span creation and updates as well, or collect output from the production formatter, and exercise an authenticated request carrying the cookie.

Themezv (commented):
Reviewed commit 4369f49e. All 196 existing CAS tests passed. Two temporary probes in an isolated copy reproduced the issues noted below.

apps/cas/src/sessions/http/extract.rs
Themezv: Right: `Router::nest` rewrites the request URI to the remainder before the inner router's extractors run, so under `/api` the warning saw `/` or `/me` instead of the path the client sent.

Fixed in 982870c9: the extractor reads the path from `OriginalUri` (the extension `nest` leaves behind) and falls back to `parts.uri` outside a nested router. Both the warning and the malformed-cookie `debug` use it. The extractor tests now mount their route under `/api` and assert `path="/api/whoami"` exactly, so a regression to the stripped path fails them.

apps/cas/src/testing.rs
Themezv: Agreed, and this was the more important of the two: a search of event fields alone would have called a log clean while `tracing_subscriber::fmt` printed the request span's fields next to every line of the request.

Fixed in 982870c9. `CaptureLayer` now implements `on_new_span` (the span's attributes at creation, rendered like an event's fields) and `on_record` (fields recorded later), so the capture holds what the production formatter would print. A unit test on the capture itself creates a span with a field, records another after entering it and emits an event inside, and asserts all three values are captured. A new transport test sends an authenticated `GET /me` with the cookie through `with_middleware`, asserts the request span was captured with its URI, and that the token appears in no span or event, so enabling headers on `DefaultMakeSpan` fails it. The doc comment and `docs/TESTS.md` say why spans are part of the capture.

PR #706, opened 2026-09-12
feat(cas): __Host- prefix for the session cookie on https origins (#700)

### LINKED ISSUE #700: cas: __Host- prefix for the session cookie on https origins

Themezv:
Closes #700.

## Changes

- **The cookie name follows the `Secure` setting.** `CookieSettings::name()` returns `__Host-cas_session` when `secure` (an https `CAS_ORIGIN`) and `cas_session` otherwise. The prefix makes the browser refuse the cookie unless it is `Secure`, has no `Domain` and `Path=/`, which is how the cookie is set already (ADR 0004 (d)); it closes cookie tossing from sibling subdomains. The plain-http development origin has no `Secure` (Safari refuses it over http even from localhost), so it has no prefix either.
- **One name for every reader.** `SESSION_COOKIE` is gone; the `Authenticated` extractor, `logout`, the removal cookie and the test helpers all read `state.cookies.name()`. A cookie under the other deployment's name does not authenticate.
- `testing::test_state_with_cookies(pool, CookieSettings)` and `test_cookies()` for tests that need the secure setting; `session_cookie(response, name)` takes the name to look for.
- ADR 0004 (d) amended, with a Status line as for #697.

## Tests

Both wire names and the `Secure` mapping; the https cookie is `__Host-cas_session` with `Secure`, `Path=/`, no `Domain`, and its dev counterpart; the removal cookie under both settings; through the router, a secure-settings logout clears `__Host-cas_session` and really ends the session, and a cookie under the other name does not authenticate while the deployment's own does. The renewal layer needed no change.

`cargo test`: 167 passed. `cargo clippy --all-targets -- -D warnings`: clean. No SQL changes.

Conflicts with #699's PR are small: the last Consequences bullet of ADR 0004 and the `testing::session_cookie` signature.

### DIFF
--- a/apps/cas/src/sessions/http/extract.rs
+++ b/apps/cas/src/sessions/http/extract.rs
@@ -14,7 +14,6 @@
 
 use axum::extract::{FromRef, FromRequestParts};
 use axum::http::request::Parts;
-use axum_extra::extract::CookieJar;
 
 use crate::http::ApiState;
 use crate::http::error::ApiError;
@@ -39,16 +38,12 @@ where
 
     async fn from_request_parts(parts: &mut Parts, state: &S) -> Result<Self, Self::Rejection> {
         let state = ApiState::from_ref(state);
-        // Reading the cookie header cannot fail; a missing or unparsable
-        // header is an empty jar.
-        let jar = CookieJar::from_request_parts(parts, &state)
-            .await
-            .unwrap_or_default();
 
         // No cookie is the ordinary state of a browser that has not signed
         // in, and of every request to a protected route from one: nothing
-        // happened, so nothing is logged.
-        let Some(cookie) = jar.get(state.cookies.name()) else {
+        // happened, so nothing is logged. The name is matched on the wire,
+        // undecoded (see `CookieSettings::presented`).
+        let Some(value) = state.cookies.presented(&parts.headers) else {
             return Err(unauthenticated());
         };
 
@@ -56,7 +51,7 @@ where
         // session — a truncated cookie, another service's cookie under the
         // same name — and says nothing about this service's sessions, so it
         // is only worth a `debug`.
-        let Some(token) = SessionToken::parse(cookie.value()) else {
+        let Some(token) = SessionToken::parse(&value) else {
             tracing::debug!(
                 path = original_path(&parts.extensions, &parts.uri),
                 "session cookie is not a well-formed token"
--- a/apps/cas/src/sessions/http/extract.rs
+++ b/apps/cas/src/sessions/http/extract.rs
@@ -19,7 +19,6 @@ use axum_extra::extract::CookieJar;
 use crate::http::ApiState;
 use crate::http::error::ApiError;
 use crate::http::extract::original_path;
-use crate::sessions::http::cookie::SESSION_COOKIE;
 use crate::sessions::http::renewal::{self, RenewalSlot};
 use crate::sessions::{Authenticated, Renewal, SessionToken};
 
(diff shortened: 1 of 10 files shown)


apps/cas/src/sessions/http/extract.rs
```diff
@@ -49,7 +48,7 @@ where
         // No cookie is the ordinary state of a browser that has not signed
         // in, and of every request to a protected route from one: nothing
         // happened, so nothing is logged.
-        let Some(cookie) = jar.get(SESSION_COOKIE) else {
+        let Some(cookie) = jar.get(state.cookies.name()) else {
```
Themezv: **[P1] Match the raw cookie name before percent-decoding**

`axum_extra::extract::CookieJar` uses `Cookie::parse_encoded`, which percent-decodes cookie names. Consequently, this lookup also accepts `%5F%5FHost-cas_session=<token>` as `__Host-cas_session`.

A compromised sibling such as `evil.example.com` can set:

```http
Set-Cookie: %5F%5FHost-cas_session=<attacker-session-token>; Domain=example.com; Path=/; Secure
```

The browser preserves the encoded name, so it does not apply the `__Host-` restriction and can send this domain cookie to `cas.example.com`. The server then decodes the name and authenticates the supplied session. This allows an attacker to inject their own valid session into a victim's browser, bypassing the cookie-tossing protection this PR intends to add. It does not require guessing another user's session token.

I reproduced the server-side behavior on this PR commit with a router test using `CookieSettings { secure: true }`: a request to `/me` carrying only `%5F%5FHost-cas_session=<valid-token>` returned **200**, while the regression assertion expected **401**. All 203 existing tests passed. Browser behavior was checked against the [cookie specification, section 5.6](https://www.ietf.org/archive/id/draft-ietf-httpbis-rfc6265bis-22.html#section-5.6), which explicitly prohibits decoding percent-encoded sequences; I did not run a browser reproduction.

Please match the literal wire name without percent-decoding it in authentication and logout, and add a regression test rejecting encoded aliases.

Themezv: Confirmed, and thank you: `axum_extra::extract::CookieJar` builds every cookie with `Cookie::parse_encoded`, so the lookup accepted the percent-encoded spelling, and a browser would neither decode it nor hold it to the `__Host-` rules. That is exactly the tossing path the prefix is supposed to close.

Fixed in bb1654b5. `CookieSettings::presented(&HeaderMap)` reads the `Cookie` header(s) directly with `Cookie::split_parse` (no decoding) and compares the name byte for byte; the `Authenticated` extractor and `logout` both go through it, and the jar is only used to emit the removal cookie on the way out. Regression coverage: unit tests on `presented` for exact matches, several `Cookie` headers, and the aliases `%5F%5FHost-cas_session`, `__Host-cas%5Fsession`, `%5f%5fHost-cas_session` and `cas%5Fsession`; a router test under both `CookieSettings` sends a valid token under an alias and asserts `/me` is 401 and a logout under the alias does not end the real session. ADR 0004 (d) now states that the name is matched on the wire and why.

Issue #723, opened 2026-09-12
cas-frontend: e2e with a virtual authenticator


Themezv:
Part of #671, last. End-to-end tests against a real CAS and Postgres.

- Playwright in `apps/cas-frontend/e2e`, Chromium with a virtual authenticator over CDP (`WebAuthn.enable`, `WebAuthn.addVirtualAuthenticator` with `hasResidentKey`, `hasUserVerification`, `isUserVerified`).
- Scenarios: create account → dashboard; sign out → sign in with the button; conditional mediation; rename and delete a passkey; last-passkey guard.
- A job in `.github/workflows/cas-frontend-pr.yml`: Postgres service as in `cas-pr.yml`, `cas-migrate`, the CAS binary, vite dev server via Playwright's `webServer`.

**Acceptance criteria**
- The scenarios pass locally and in CI on a PR.

Issue #715, opened 2026-09-12
cas-frontend: readable ceremony and API errors, remove tests/client


Themezv:
Part of #670, last. One place that turns every failure into a Russian sentence, and the old throwaway client goes away.

- Map `ApiError` codes (`unauthenticated`, `invalid_display_name`, `invalid_credential`, `cross_site_request`, `not_found`, 5xx, network failure) and WebAuthn `DOMException` names (`NotAllowedError` = cancelled or timed out, `InvalidStateError` = this passkey already exists here, `NotSupportedError` / `ConstraintError` = the authenticator cannot make a discoverable credential, `SecurityError` = wrong origin / rp id) to user messages in `src/shared/errors` (or similar). Unit tests on the mapping.
- Screens use the mapping; no raw `message` reaches the UI.
- Remove `apps/cas/tests/client`.

**Acceptance criteria**
- Every listed code and exception has a message and a test.
- `apps/cas/tests/client` is gone and nothing references it.

Issue #712, opened 2026-09-12
cas-frontend: sign-out and auth gate


Themezv:
Part of #670, after create account (a session has to exist to test this).

- `POST /api/logout` in `entities/session`; a sign-out button on the dashboard, `Clear-Site-Data` does the rest.
- Root loader calls `/api/me`: `unauthenticated` redirects to `/sign-in`; a signed-in visitor of `/sign-in` or `/create-account` is redirected to `/`.
- After sign-out the router revalidates and lands on `/sign-in`.

**Acceptance criteria**
- Opening `/` signed out lands on `/sign-in`; opening `/sign-in` signed in lands on `/`.
- Sign out ends the session (a following `/api/me` is 401) and shows the sign-in screen.

PR #726, opened 2026-09-15
feat(cas-frontend): create account (#711)

### LINKED ISSUE #711: cas-frontend: create account
Part of #670. The first vertical slice that proves the whole pipeline: display name → passkey ceremony → signed-in dashboard.

- `/create-account`: a form with one field (display name, required, max 64) on `useActionState`; `POST /api/webauthn/register-options`, `startRegistration` from `@simplewebauthn/browser`, `POST /api/webauthn/verify-registration`, then navigate to `/`.
- `entities/session`: `getMe` and the `MeResponse` type; the dashboard shows the display name from `/api/me`.
- Minimal error handling: `invalid_display_name` and a cancelled ceremony show a readable Russian message. The full mapping is a later ticket.
- Add `@simplewebauthn/browser` (catalog) to the app.

**Acceptance criteria**
- A new account is created with a real passkey (Touch ID / iCloud Keychain, or Chrome's virtual authenticator) and the dashboard greets it by name.
- A rejected display name and a cancelled ceremony leave the form usable with a message, not a raw exception.

Themezv:
Closes #711.

## What

The first vertical slice through the whole pipeline: display name → passkey ceremony → signed-in dashboard.

- **`/create-account`**: one field on `useActionState`. The client checks emptiness and length (64 code points, counted as the server counts them); `POST /api/webauthn/register-options`, `startRegistration` from `@simplewebauthn/browser`, `POST /api/webauthn/verify-registration`, then navigate to `/`. Copy and states follow `docs/DESIGN.md`.
- **Errors, minimal as the ticket asks**: the server's `invalid_display_name` lands under the field and the name stays; a cancelled or timed-out ceremony (`NotAllowedError`) is an alert above the form with the button reading "Попробовать ещё раз". Anything else is a generic alert; the full mapping is #715. `NotAllowedError` is matched by name: `@simplewebauthn/browser` rethrows the `DOMException` as its own `WebAuthnError` that keeps the name.
- **`entities/session`**: the `Me` type, `getMe`, `registerWithPasskey`.
- **Dashboard**: the route loads `/api/me` and greets by name. The root route gets the loading screen as `HydrateFallback` and an error screen; without a session the dashboard's loader is a 401 and the error screen offers sign-in, until #712 adds the redirect. Pages draw their own `Screen`, so the scaffold's root layout goes.
- `@simplewebauthn/browser` (catalog) joins the app; `@testing-library/user-event` joins the catalog for the page spec.

## Checks

- `pnpm ts-check`, `pnpm lint:check`, `pnpm fmt:check`, `pnpm build`: pass.
- Unit tests (15): the validation, the failure mapping (raw `DOMException` and the simplewebauthn wrapper), and the page's action: trimmed name reaches the ceremony and the dashboard follows; empty name never hits the server; server verdict under the field with the name kept; cancelled ceremony as an alert above a still usable form.
- End to end against a local CAS with Chrome's virtual authenticator over CDP (an ad-hoc Playwright script; the real e2e job is #723): empty name rejected; a zero-width space in the name rejected by the server and shown under the field; user verification refused → "Создание отменено", form still usable; then the same form creates the account, the authenticator holds one discoverable credential, `/api/me` names it and the dashboard greets by name; a fresh browser on `/` gets the error screen with "Войти".


apps/cas-frontend/src/pages/create-account/validateDisplayName.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Character rules stay on the server, which answers `invalid_display_name`.
+ * Returns the message to show under the field, or `null` for a name worth
+ * sending.
+ */
+export const validateDisplayName = (name: string): string | null => {
+  if (name.length === 0) {
+    return messages.empty
+  }
+  if ([...name].length > MAX_DISPLAY_NAME_LENGTH) {
+    return messages.tooLong
```
Themezv, line 39: Could we normalize the display name before checking its length, using the same order as CAS?

The backend applies NFC normalization and collapses spaces before enforcing the 64-code-point limit. Here, `'e\u0301'.repeat(33)` is counted as 66 code points and rejected, although NFC turns it into 33 `é` characters that the server accepts. I reproduced this in a page test: the validation error prevents `registerWithPasskey` from being called.

No extra dependency or character mapping table is needed; JavaScript provides NFC normalization and Unicode character classes:

```ts
const normalizeDisplayName = (name: string) =>
  name
    .replace(/\p{Zs}/gu, ' ')
    .normalize('NFC')
    .replace(/ +/g, ' ')
    .replace(/^ +| +$/g, '')
```

We could use this in place of the current `.trim()` in the form action, then validate and submit the normalized value. Trimming ASCII spaces here also matches CAS more closely than `.trim()`, which removes some characters the server intentionally rejects. The more involved character validation can stay on the server.

A regression test for decomposed characters near the length limit would cover this case.

Themezv: Done in 46c558a1: `normalizeDisplayName` maps every Zs space to ASCII space, applies NFC, then trims and collapses ASCII spaces only, in the same order as `sanitize_label`. The form action validates and sends the normalised value. Tests: decomposed `é` × 33 is accepted and reaches `registerWithPasskey` in NFC, × 65 is too long; a zero-width space at the edge is no longer trimmed, so the server's `invalid_display_name` still applies to it.

Themezv:
Follow-up on where error handling lives: `shared/errors` now holds only the `Failure` shape and the generic fallback. The create-account copy ("Создание отменено") sits in the page, and `isCeremonyCancelled` next to the ceremony that throws it, in `entities/session`. The shared `toFailure(error, context)` with its per-context title map is gone. `docs/CODE.md` states the rule; note it narrows the wording of #715 ("one place in `src/shared/errors`") to: generic in shared, screen-specific next to the screen.

Issue #671, opened 2026-08-30
cas-frontend: dashboard with passkey management


Themezv:
Signed-in dashboard: list passkeys (name, created, last used), add/rename/delete, nudge to add a second passkey (the v1 recovery story), optional email field (stored unverified; verification comes with a later milestone). UI copy in Russian.

Depends on #668, #670.

**Acceptance criteria**
- Passkey CRUD works end-to-end against the management API.
- Last-passkey deletion is blocked in UI with an explanation.
- Email can be saved and cleared; no verification flow.

PR #732, opened 2026-09-17
feat(cas-frontend): rename a passkey (#717)

### LINKED ISSUE #717: cas-frontend: rename a passkey
Part of #671, after the list.

- Inline rename per row: a form action on `PATCH /api/passkeys/{id}` with `useOptimistic` on the list; `invalid_passkey_name` (empty, over 64, invisible characters) shows a message and keeps the old name.
- Revalidate the dashboard after success.

**Acceptance criteria**
- Renaming updates the row immediately and survives a reload.
- A rejected name rolls back and explains why.

Themezv:
Closes #717. Part of #671, after #731.

## What

- **`entities/passkey`**: `renamePasskey(id, name)` over `PATCH /api/passkeys/{id}`, answering the passkey as it now is.
- **`PasskeyRow`**: a 44px rename button on every row. It turns the row into the field "Название" with the current name, "Сохранить" and "Отмена" (Escape cancels too); focus goes back to the button when the form closes. The form is a React action in its own `Editor`, mounted for one editing session so the next one starts clean.
- **`DashboardPage`**: `useOptimistic` on the list. The row reads the new name the moment "Сохранить" is pressed, the form is gone, and the action stays pending until the loader has the name from the server, so nothing flickers back. A `passkey_not_found` on rename (deleted in another tab) reloads the list instead of complaining: the row goes.
- **Rejections**, in the row's own words under the field, never the raw message: an empty name and one over 64 code points never leave the browser ("Введите название.", "Слишком длинное название, максимум 64 символа."); `invalid_passkey_name` from the server brings the form back with the rejected value and "Название содержит недопустимые символы." while the list keeps the old name; anything else is "Не получилось переименовать. Попробуйте ещё раз через минуту."
- **`shared/lib/label.ts`**: the name rules the create-account screen repeated from `apps/cas/src/shared/label.rs` (`normalizeLabel`, `labelProblem`, `MAX_LABEL_LENGTH`) move here now that a second screen needs them, as CODE.md says. `validateDisplayName` keeps that screen's words and maps onto them.
- DESIGN.md carries the rename state and its copy; LAYOUT.md lists `shared/lib/`.

## Tests

- `DashboardPage.spec.tsx`, rename: the new name shows before the server answers and survives the reload, sent normalised; a rejected name comes back in the field with the reason and the list keeps the old one; empty and over-long names are not sent; an unchanged name closes the form without a request; a network failure keeps the form with its own words; a passkey deleted elsewhere leaves the list; cancel puts focus back on the button.
- `label.spec.ts` takes over the normalisation and counting cases from `validateDisplayName.spec.ts`, which now covers only the mapping to that screen's words.
- In the browser against a running CAS: `/` without a session still lands on `/sign-in` with no console errors beyond the gate's 401. The rename itself needs a signed-in browser: please open http://localhost:5173 with your passkey, the dev server on 5173 serves this branch.

`pnpm lint:check`, `pnpm fmt:check`, `pnpm ts-check`, `pnpm test:ci` (77 tests) and `pnpm build` pass.


apps/cas-frontend/src/pages/dashboard/PasskeyRow.tsx
```diff
@@ -1,20 +1,153 @@
...
+/** One passkey in the list: the key chip, the name, when it was made and last used, and the way to rename it. Delete comes with #718. */
+export const PasskeyRow = ({ passkey, onRename }: PasskeyRowProps) => {
+  const [editing, setEditing] = useState(false)
+  const renameButton = useRef<HTMLButtonElement>(null)
+  const wasEditing = useRef(false)
+
+  // Focus goes back to where the editing began, after a save and a cancel alike; not on first render.
+  useEffect(() => {
+    if (wasEditing.current && !editing) {
+      renameButton.current?.focus()
```
Themezv, line 39: [P2] Preserve focus when another row is being edited

Save one passkey, then start editing another while the first request is pending. When the first save completes, this effect unconditionally focuses its rename button, moving focus out of the second input and interrupting typing. I reproduced this with a component test using a deferred rename response. Please restore focus after an asynchronous save only if the user has not moved to another control during the request.

PR #733, opened 2026-09-19
feat(cas-frontend): delete a passkey with the last-passkey guard (#718)

### LINKED ISSUE #718: cas-frontend: delete a passkey with the last-passkey guard
Part of #671, after rename.

- Delete per row behind a `<dialog>` confirmation; `DELETE /api/passkeys/{id}` with `useOptimistic`.
- The last passkey cannot be deleted: the control is disabled with an explanation in the UI, and `409 last_passkey` from the API (a race) shows the same message.
- `passkey_not_found` (deleted in another tab) revalidates the list.

**Acceptance criteria**
- With two passkeys, deleting one removes it and leaves the other's delete disabled with the explanation.
- The 409 path is covered by a unit test on the action.

Themezv:
Closes #718.

## What

- `entities/passkey`: `deletePasskey` over `DELETE /api/passkeys/{id}`.
- A passkey row gets a 44px delete button next to rename. It opens a modal `<dialog>` sheet at the bottom of the dimmed page: "Удалить пасскей «…»?", danger "Удалить" with the trash icon, secondary "Отмена". Escape and a tap on the backdrop cancel; focus starts on "Отмена".
- Confirming is a form action: the sheet closes, the row leaves the list at once with `useOptimistic`, the route revalidates once the server has answered, so the row never flickers back.
- **Last-passkey guard**: with one passkey the delete button is disabled (hint colour) with "Единственный пасскей нельзя удалить: сначала добавьте второй." under the row, linked by `aria-describedby`.
- `409 last_passkey` (a race with another tab): the row comes back with the same sentence and the list reloads, so the button is off from then on.
- Any other failure: the row comes back with "Не получилось удалить. Попробуйте ещё раз через минуту." in its own words, never the raw message.
- `passkey_not_found` (deleted in another tab): the list reloads and the row goes with no message.
- `docs/DESIGN.md` documents the delete states; stories cover the sheet, the last-passkey row and the failed row.

## Tests

jsdom has no modal dialogs, so the spec polyfills `showModal`/`close` (with the `close` event and focus return). New cases: the sheet and optimistic removal, cancel with focus back on the button, the disabled last passkey with its note, both 409 outcomes (list caught up / not), a network failure, a passkey deleted elsewhere.

`pnpm ts-check`, `pnpm test:ci` (85 passed), `pnpm lint:check`, `pnpm fmt:check` are clean. The sheet, the last-passkey row and the failed row were checked visually in Storybook at phone and desktop widths; backdrop tap closes the sheet.


apps/cas-frontend/src/pages/dashboard/DashboardPage.tsx
```diff
@@ -1,31 +1,39 @@
...
+/** The account by name, its passkeys, and the way out. Adding passkeys and the email come with #720 and on. */
 export const DashboardPage = () => {
   const { me, passkeys } = useLoaderData<DashboardData>()
   const revalidator = useRevalidator()
-  // React shows the renamed list while the row's action runs and goes back to the loader's once it settles.
-  const [shownPasskeys, showRenamed] = useOptimistic(passkeys, withRenamed)
+  // React shows the changed list while the row's action runs and goes back to the loader's once it settles.
+  const [shownPasskeys, showChange] = useOptimistic(passkeys, withChange)
+  // The row that could not be deleted explains why once it is back in the list; the next delete starts clean.
+  const [deleteFailure, setDeleteFailure] = useState<{ id: string; reason: DeleteFailure } | null>(null)
```
Themezv: [P2] Preserve errors for overlapping deletions

With at least three passkeys, a user can confirm two deletions before either request finishes. If both requests fail, both rows return, but this single `deleteFailure` slot retains only the last error, leaving the other restored row without an explanation. I reproduced this with a regression test: after rejecting both pending requests, all three rows return but only one deletion error is rendered.

Please track failures by passkey ID, clearing only the relevant entry on retry, or serialize deletion requests so failures cannot overwrite each other.

Themezv: Fixed in f92c3c95: delete failures are now keyed by passkey id, and a retry clears only its own row's entry. Added two regression tests: two in-flight deletes that both fail keep a reason on both rows (verified it fails on the previous code), and retrying one of two failed rows takes away only that row's words.

PR #736, opened 2026-09-19
feat(cas): update the account email (#721)

### LINKED ISSUE #721: cas: update the account email
Backend, part of #671. `GET /api/me` returns `email` but nothing can set it; the dashboard needs to save and clear it (unverified, v1).

- An endpoint to set or clear the account's email (e.g. `PATCH /api/me` with `{ "email": string | null }`), `Authenticated`, syntactic validation only, stored unverified as today.
- Stable error code for an invalid address.

**Acceptance criteria**
- Setting, changing and clearing the email round-trips through `GET /api/me`.
- An invalid address is `400` with a stable code; an anonymous request is `401`.

Themezv:
Closes #721.

## Route

`PATCH /api/me` with `{ "email": string | null }`, `Authenticated`, answers `204 No Content`. The account id comes from the session, never from the body; `null` clears; the call is idempotent (the same body twice leaves the same row, clearing an absent address is a `204` too). The field is required: a body without it is `422 invalid_body`, so `{}` cannot clear an address the client never mentioned. Nothing is returned; the dashboard reads the account back through `GET /api/me`, whose shape does not change.

The write belongs to the accounts context: `accounts/http` (the router and the error mapping) and `accounts/management.rs` (`AccountManagement::set_email`) are new, and `ApiState` gains `accounts`. The transport root merges the accounts router at `/api` next to the sessions one, and axum merges the two method routers for `/me`: `GET` stays the session's question, `PATCH` is the account's write. No new prefix, so the Fetch Metadata line covers it by where `/api` is; the representative-route CSRF test lists `PATCH /api/me`.

## Validation

`Email` newtype in `accounts/email.rs` (nutype, like `DisplayName`), syntax only, no new dependency. In order:

- surrounding whitespace trimmed;
- the domain (after the last `@`) lower-cased, the local part left as typed;
- not empty; at most 254 characters (RFC 5321's forward-path limit);
- no whitespace, control, format (invisible), separator, surrogate, private-use or unassigned character anywhere;
- exactly one `@`, with a non-empty local part and a non-empty domain;
- the domain is non-empty dot-separated labels (`ada@example.` and `ada@.com` are refused).

Non-ASCII and dotless domains pass: the rules are about shape, deliverability is the verification flow's job (M4).

## Error codes

- `400 invalid_email`, message `Invalid email: <rule>` (one `EmailError` variant per rule).
- `401 unauthenticated` for an anonymous request, from the extractor before the body is read.
- `403 cross_site_request` from the CSRF layer.
- `422 invalid_body` for a body without the field.
- `404 account_not_found` for the one race the write can hit (the account deleted under a live session); covered by a service test since the cascade takes the session with it.

## Storage and ADR

No migration: `accounts.email` is already nullable `text` with no unique constraint, and it stays that way (the accounts migration records why an unverified address must not be unique; a test now states that two accounts may share one). The `Email` guards the write; `Account::email` stays `Option<String>` and the reader does not re-validate, so a stricter rule later cannot lock an account out of `/me`. The change is logged at `info` with the account id and whether it was set or cleared, never the address.

New [ADR 0007 "The account email"](apps/cas/docs/adr/0007-account-email.md) records the route and its ownership (a–b), the rules (c), storage (d), the not-found race (e) and logging (f). One query joined the `.sqlx` offline cache.

## Tests

- `accounts/email.rs`: ordinary addresses kept (ASCII, plus-addressing, Cyrillic, CJK, dotless), trim and domain-only lower-casing, empty, over the limit, spaces/control/invisible characters, `@` count and empty parts, empty domain labels, the order of the checks, deserialising validates.
- `accounts/http`: set, change and clear round-trip through `GET /api/me`; normalisation of a valid address; idempotence; nine invalid shapes are `400 invalid_email` and leave the stored address alone; a body without the field is `422`; anonymous is `401` whatever the body.
- `accounts/management.rs`: set/change/clear, clearing twice, unknown account, the log line carries the id and not the address.
- `accounts/repository.rs`: the write and the clear, an unknown id, non-uniqueness.
- `http/mod.rs`: `PATCH /api/me` is reachable under the prefix (anonymous `401`) and cross-site is `403`.

Run from `apps/cas` against a throwaway database (`cas_721`) on the dev Postgres, mirroring `cas-pr.yml`:

- `cargo fmt --check` — clean
- `SQLX_OFFLINE=true cargo clippy --all-targets --all-features` — 0 warnings
- `SQLX_OFFLINE=true DATABASE_URL=postgres://cas:cas@localhost:5434/cas_721 cargo test` — 247 passed, 0 failed
- `DATABASE_URL=postgres://cas:cas@localhost:5434/cas_721 cargo sqlx prepare --check` — fresh
- `check-migration-order.sh` not applicable: no migration added.


apps/cas/src/accounts/email.rs
```diff
@@ -0,0 +1,235 @@
...
+    if domain.contains('@') {
+        return Err(EmailError::MultipleAt);
+    }
+    if local.is_empty() {
+        return Err(EmailError::EmptyLocalPart);
+    }
+    if domain.is_empty() {
+        return Err(EmailError::EmptyDomain);
+    }
+    if domain.split('.').any(str::is_empty) {
```
Themezv, line 136: [P2] Validate domain characters before accepting an address

The non-empty-label check accepts `ada@example.com/` and `ada@exam,ple.com`; I reproduced both as successful `Email::try_new` results. An authenticated PATCH therefore stores these malformed addresses and returns 204 instead of `400 invalid_email`, potentially replacing a previously valid value. This is a syntax check that can be performed without testing mailbox existence: neither slash nor comma is valid in these domain labels under [RFC 5321 section 4.1.2](https://www.rfc-editor.org/rfc/rfc5321#section-4.1.2). Please validate domain-label syntax while preserving the intended internationalized-domain support, and add regression cases. ADR 0007 explicitly chooses the current loose rules, so the corresponding policy text should change with the validator.

apps/cas/src/accounts/email.rs
```diff
@@ -0,0 +1,235 @@
...
+        )
+}
+
+/// The syntactic rules, in the order a user would want to hear about them.
+/// Runs on the sanitised value: trimmed, domain lower-cased.
+fn validate_email(value: &str) -> Result<(), EmailError> {
+    if value.is_empty() {
+        return Err(EmailError::Empty);
+    }
+    if value.chars().count() > MAX_EMAIL_LENGTH {
```
Themezv: [P2] Apply the SMTP-derived length limit in bytes

`chars().count()` does not enforce the 254-octet limit cited as the rationale for this bound. I reproduced acceptance of an address consisting of 32 `é` characters, `@`, and three dot-separated 63-character ASCII labels: 224 Unicode scalar values but 256 UTF-8 bytes. The local part itself is only 64 bytes, so this specifically exposes the missing total-byte bound. [RFC 5321 section 4.5.3.1](https://www.rfc-editor.org/rfc/rfc5321#section-4.5.3.1) explicitly distinguishes octets from character counts. If this bound is intended to enforce that SMTP limit, use a byte-length check and add a Unicode boundary test, updating the error message and ADR accordingly. If 254 characters is intentionally only a product/storage limit, document that distinction and remove the claim that it enforces the SMTP limit.

Themezv (commented):
Two validation findings below. The implementation follows the deliberately simplified rules in ADR 0007; these comments propose tightening those rules to match the stated syntactic-validation goal and SMTP length rationale. Since email is currently only stored unverified, neither finding demonstrates a current sign-in or recovery failure.

Validation: all 247 CAS library tests passed using this revision's workspace lockfile and isolated SQLx test databases. Additional executable probes reproduced both cases. No repository changes were made.

apps/cas/src/accounts/email.rs
Themezv: Fixed in 57d2cbdc: every label must now be letters, digits and hyphens (RFC 5321 §4.1.2), not starting or ending with a hyphen; letters and digits are read across every script and combining marks are allowed, so `пример.рф` and `मेल.भारत` still pass while `example.com/`, `exam,ple.com`, `exa_mple.com`, `-example.com`, `[127.0.0.1]` are `400 invalid_email` with the new `InvalidDomainLabel` rule. ADR 0007 (c) updated; regression cases added.

apps/cas/src/accounts/email.rs
Themezv: Fixed in 57d2cbdc: the bound is now `value.len()` (UTF-8 octets), the constant's doc and the error message say bytes, ADR 0007 (c) says octets as RFC 5321 counts them. Boundary test added: 32 `é` + `@` + three 63-char labels (224 characters, 256 bytes) is rejected, and the 254-byte variant with 31 `é` is accepted.

Themezv:
По обсуждению в ревью: существующая серверная валидация остаётся. Для клиентской проверки предлагаю использовать нативную валидацию `type="email"` в связанном frontend PR #737.

В `EmailSection.tsx` уже выставлены `type="email"`, `inputMode="email"` и `autoComplete="email"`, но `<form noValidate>` отключает браузерную проверку при отправке. Собственная проверка сейчас проверяет только пустоту и наличие `@`, поэтому, например, `ada,lovelace@example.com` проходит её; текущая серверная проверка тоже принимает этот пример.

Предложение для #737:
- Убрать `noValidate` с формы, чтобы браузер проверял адрес при сохранении.
- Добавить `formNoValidate` на кнопку «Удалить», чтобы некорректный текст в поле не мешал очистить сохранённую почту.
- Сохранить обработку серверного `invalid_email` для остальных ограничений.

Если нужны именно собственные тексты ошибок, можно вместо браузерного сообщения использовать `input.validity.typeMismatch`, сохранив нативную проверку формата.

Themezv:
Applied to #737 in 433b5eec, the second variant: `noValidate` stays so the section keeps its own words, and the action reads `validity.typeMismatch` from the `type="email"` field instead of the bare `@` check, so `ada,lovelace@example.com`, `exa_mple.com`, `-example.com` never leave the browser. A clear ("Удалить") sends `null` whatever the field holds, which is what `formNoValidate` would have done. Server `invalid_email` handling unchanged; the spec's server-rejection case now uses an over-the-cap address, the one rule the browser does not count.

PR #737, opened 2026-09-19
feat(cas-frontend): email on the dashboard (#722)

### LINKED ISSUE #722: cas-frontend: email on the dashboard
Part of #671, blocked by the backend ticket for updating the email (linked as a sibling sub-issue).

- Email field on the dashboard: save and clear via a form action; the nudge to add one ("recovery arrives later, this is where it will go").
- No verification flow; the UI says the address is unverified.

**Acceptance criteria**
- Email can be saved, changed and cleared; the value survives a reload.
- An invalid address shows a message and keeps the field editable.

Themezv:
Closes #722.

## The section

"Почта" under the passkeys, a `Section` with the mail chip and one row (`EmailSection.tsx`). The empty state is the whole nudge for the email, as `docs/DESIGN.md` has it; there is no second accent card.

- **Empty**: "Не указана" / "Понадобится для восстановления, когда оно появится. Пока без подтверждения.", with "Добавить" (plus icon, `aria-label="Добавить почту"`) in the title row where the passkeys' "Добавить" is.
- **Set**: the address, "Не подтверждена. Подтверждение появится позже." and the 44px pencil (`aria-label="Изменить почту"`). No verification flow in v1; the row says so.
- **Editing**: the row becomes the field "Почта" (`type="email"`, `inputMode="email"`, `autoComplete="email"`, `noValidate` on the form so the browser's own check does not speak first) holding the current address, the secondary "Сохранить" and "Отмена", and, when an address is set, the danger-coloured inline "Удалить" with the trash icon on the right, a second submit of the same form that sends `null`. Escape cancels. Focus goes back to the control that opened the form (the pencil after a save from empty, "Добавить" after a clear) unless the user has moved on.
- **Pending**: the row shows the new address (or the empty state after a clear) at once through `useOptimistic` on the page, with the spinner in place of the pencil, and the action stays pending until `revalidate()` has read the account back through `GET /api/me`. So the row ends up with what the server stored (the domain lower-cased: "Ada@Mems.fun" comes back as "Ada@mems.fun") and that is what survives a reload.

## API and error mapping

`updateEmail(email: string | null)` in `entities/session/api.ts`, next to `getMe`: `PATCH /api/me` with `{ email }`, a 204 (#736).

- Trimmed client-side; a submit that did not change the address closes the form without a request.
- Empty: "Введите адрес.", no request.
- No `@`: "Похоже, это не адрес почты.", no request.
- `400 invalid_email` (spaces or invisible characters, a second `@`, an empty part, a broken domain, over 254 bytes): the form comes back with the rejected value and "Проверьте адрес: в нём ошибка или недопустимые символы."; the row keeps what it had; never the raw message.
- Anything else (`account_not_found`, no network, an outage): "Не получилось сохранить почту. Попробуйте ещё раз через минуту." under the field, the value kept; a failed clear brings the form back with the address and the same sentence.
- `401 unauthenticated`: no message; the page revalidates and the loader sends the browser to `/sign-in`.

## Docs and stories

- `docs/DESIGN.md`: the section's layout in the Dashboard list and an "Email" state block with the copy above; the "email comes later" remark is gone from the page's doc comment.
- `EmailSection.stories.tsx` ("CAS / Email"): Empty, Set, Editing, Adding, Invalid, Failed. The last four reach their state through a `play` function, so the editor stays internal.

## Tests

`DashboardPage.spec.tsx`, a new `email` block (15 tests, `updateEmail` mocked, the real loader in `createMemoryRouter`): the two states; save from empty (address visible at once, trimmed body, the pencil waits, the loader's lower-cased domain after the reload, focus on the pencil); change; clear from "Удалить" (sends `null`, empty state at once, "Добавить" back and focused); no clear control while empty; cancel and Escape (no request, focus back); an unchanged address (no request); empty submit; a value without `@`; `invalid_email` keeps the field with the value and the message and never the raw one; a generic failure and a failed clear in the row's words; `unauthenticated` lands on sign-in.

From `apps/cas-frontend`: `pnpm ts-check` clean; `pnpm test:ci` 10 files, 114 passed. From the root: `pnpm lint:check` clean; `pnpm fmt:check` clean. Checked in Storybook (`pnpm storybook`, "CAS / Email"): all six stories render, the play functions reach the editing and error states, no console errors.


Themezv:
Follow-up from the review of #736 landed in 433b5eec: the shape of the address is now judged by the browser's own `type="email"` check (`validity.typeMismatch`) read inside the action, with this section's copy; a clear sends `null` regardless of the field. Tests: 119 passed, lint/fmt/ts-check clean.

PR #738, opened 2026-09-19
test(cas-frontend): e2e with a virtual authenticator (#723)

### LINKED ISSUE #723: cas-frontend: e2e with a virtual authenticator

Themezv:
Closes #723.

End-to-end tests for `apps/cas-frontend` against a real CAS and Postgres, with the passkeys on Chromium's virtual authenticator. The last ticket of #671.

## The fixture

`e2e/fixtures.ts` opens a CDP session per test, `WebAuthn.enable`, and attaches one virtual authenticator before the test starts: `ctap2`, `transport: 'internal'`, `hasResidentKey`, `hasUserVerification`, `isUserVerified`, `automaticPresenceSimulation`. `authenticators.add()` / `.remove()` stand for another device (a second passkey is a swap, not an addition: with two attached authenticators a registration fails as `InvalidStateError` as soon as either holds an excluded credential, as it would with real ones). The `account` fixture creates a fresh account with a unique display name and lands on the dashboard, so every test is independent and the suite runs fully parallel against one CAS.

Conditional mediation: with the virtual environment on and its UI off, Chromium answers the sign-in screen's autofill offer with the discoverable credential by itself, without a click. The autofill scenario relies on that and records the `mediation` of every `navigator.credentials.get()` to be sure it was the offer, not the button, that signed in. The flip side is that the button cannot be tested while the offer stands, so the button scenarios run with the option `autofill: false`: a browser without conditional mediation (`isConditionalMediationAvailable` answers `false`), the one the button is the fallback for in DESIGN.md.

## Scenarios (`e2e/*.e2e.ts`, 9 tests)

- create-account: `/` → gate → `/sign-in` → "Создать" → name → dashboard with the name, the nudge, one "Passkey" row never used, delete disabled with the last-passkey note; the authenticator holds one resident credential for `localhost`; a signed-in browser is sent away from `/sign-in`. Plus: a name the server rejects (`invalid_display_name`) stays in the field with the reason.
- sign-in: sign out → sign in with the button (`autofill: false`), "Использован сегодня" on the row; a device without a passkey ends with "Вход отменён" and stays on sign-in; sign in from the autofill offer without pressing anything (conditional mediation).
- passkeys: the only passkey cannot be deleted (disabled, described by the note, no sheet); rename, the name survives a reload, "Отмена" keeps the old one; add a second passkey (the same device is refused with "На этом устройстве уже есть пасскей", another device succeeds: nudge gone, delete enabled on both rows); delete through the sheet ("Отмена" and Escape leave the list, confirm removes the row, the server's 204 is awaited before the reload) and the guard comes back on the last one.

## CI

A new `e2e` job in `cas-frontend-pr.yml`: the Postgres service from `cas-pr.yml`, `dtolnay/rust-toolchain` + `Swatinem/rust-cache` (pinned by SHA), `cargo build -p cas --bins` with `SQLX_OFFLINE`, `cas-migrate`, CAS started in the background with its defaults (`:3000`, RP id `localhost`, origin `http://localhost:5173`), `playwright install --with-deps chromium`, `wait-on` the health endpoint, `pnpm test:e2e` (vite is Playwright's `webServer`), the report, traces and CAS log uploaded on failure. The workflow's path filter now includes `apps/cas/**`, `Cargo.toml` and `Cargo.lock`, since the suite tests the real backend; the three existing frontend jobs run on those changes too, they are cheap.

## Locally

See `docs/TESTS.md`: Postgres via `apps/cas/docker-compose.yml`, a database of its own, `cargo run -p cas --bin cas-migrate`, `cargo run -p cas`, then `pnpm test:e2e` from `apps/cas-frontend`. `CAS_API_PROXY_TARGET` points at a CAS elsewhere; `CAS_FRONTEND_PORT` moves vite when 5173 is taken (CAS then needs `CAS_ORIGIN` / `CAS_CORS_ORIGINS` to match). Ran 5× locally: 45/45 passed.

Also: `@playwright/test` added to the app from the catalog (already used by ligretto-frontend), `e2e/` and `playwright.config.ts` type-checked through the app's tsconfig, `playwright-report/` and `test-results/` ignored in the app.


apps/cas-frontend/e2e/sign-in.e2e.ts
```diff
@@ -0,0 +1,69 @@
...
+    const get = navigator.credentials.get.bind(navigator.credentials)
+    navigator.credentials.get = options => {
+      void (window as unknown as { recordMediation: (mediation: string) => void }).recordMediation(options?.mediation ?? 'none')
+      return get(options)
+    }
+  })
+
+  const displayName = uniqueName('autofill')
+  await createAccount(page, displayName)
+  await signOut(page)
```
Themezv: **[P2] Avoid waiting for the transient sign-in screen**

With autofill enabled, the virtual authenticator immediately authenticates after logout, so `signOut()` can miss `/sign-in` and its heading. The test then fails even though authentication succeeded.

Reproduced against this commit with Playwright `launchOptions: { slowMo: 150 }` and retries disabled: the other eight tests passed, but this test timed out in `signOut()` at line 7, expecting `/sign-in` while already back at `/`. CI and 40 normal-speed repetitions passed, so this is timing-dependent.

In a temporary copy, replacing this call with `await page.getByRole('button', { name: 'Выйти' }).click()` and retaining the final dashboard, last-used, and mediation assertions passed under the same slow-motion setting. Please avoid asserting the transient sign-in screen in this autofill scenario.

Themezv: Fixed in 1ba74565. The autofill test no longer asserts the transient sign-in screen: it registers `waitForResponse` for the 204 to `/api/logout` before clicking "Выйти" and awaits it, which proves the session ended without depending on how long the screen stays up, then keeps the dashboard, "Использован сегодня" and `mediation === 'conditional'` assertions. The two button tests keep `signOut`, the screen is not transient with autofill off. Reproduced your setup locally (`launchOptions: { slowMo: 150 }`, `retries: 0`): the previous revision failed 2 of 3 runs on `toHaveURL('/sign-in')` with `Received: http://localhost:5723/`; with the fix 9/9 under slowMo and the full suite 27/27 at normal speed with `--repeat-each 3`.

